PR #93, opened 2013-07-26
Include support for Pdepend and phploc

japaveh:
WIth this PR it will be possible to generate phploc reports in the build and a Pdepend graph + pyramid

Use the following commands

```
php_loc:
pdepend:
    directory: "module"
```


lakshmajim:
Hi @HMAZonderland ,
When I'm tyring to call `http://domain.com/build/pdepend/bubbles-summary.xml`, I was getting the following error as you did mentioned above
```bash
Sorry, there was a problem
Controller Build does not have action pdepend
`
How to solve this issue by adding an action and where to add that action method ?

PR #556, opened 2014-08-09
Set build path to executor in every build

### LINKED ISSUE #509: Build_path is / in Shell plugin with multiple commands and not the real path
I try to use the Shell plugin to execute some commands. But I didn't get the base path

I place $this->phpci->log($command); in PHPCI/Plugin/Shell.php on line 77 to display the command.
To test it I used the next configuration with multiple commands
```
    shell:
        - "echo $PHPCI_BUILD_PATH"
        - "echo $BUILD_PATH"
        - "echo %BUILD_PATH%"
        - "echo %build_path%"
```

The result is:
```
echo $PHPCI_BUILD_PATH
/
echo $BUILD_PATH

echo /
/
echo %build_path%
EXCEPTION: Warning: sprintf(): Too few arguments in /path/to/phpci/PHPCI/Helper/BaseCommandExecutor.php line 68
```
If I remove the last line from the config, I get the same but without that error part of this result

It use / as the BUILD_PATH and not /path/to/phpci/PHPCI/build/projectx-buildy

nonlux:
some fix #509
But it will be break?

### DIFF
--- a/PHPCI/Builder.php
+++ b/PHPCI/Builder.php
@@ -304,6 +304,7 @@ class Builder implements LoggerAwareInterface
             throw new \Exception('Could not create a working copy.');
         }
 
+        $this->commandExecutor->setBuildPath($this->buildPath);
         // Does the project's phpci.yml request verbose mode?
         if (!isset($this->config['build_settings']['verbose']) || !$this->config['build_settings']['verbose']) {
             $this->verbose = false;
--- a/PHPCI/Helper/BaseCommandExecutor.php
+++ b/PHPCI/Helper/BaseCommandExecutor.php
@@ -87,7 +87,7 @@ abstract class BaseCommandExecutor implements CommandExecutor
 
         $pipes = array();
 
-        $process = proc_open($command, $descriptorSpec, $pipes, dirname($this->buildPath), null);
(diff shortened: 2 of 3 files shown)


asterixcapri:
Every time an executeCommand() is called from a Plugin, the buildPath of the Executor is always null. So proc_open() doesn't get the proper $cwd argument.

$process = proc_open($command, $descriptorSpec, $pipes, dirname($this->buildPath), null);

This is a real problem because calling a script via Shell plugin gives a file not found error.

This patch try to fix this problem. Can you merge it or solve in another way?

Thank you

tvbeek:
@dancryer Can you show the failure for this build?

nonlux:
Sorry, I  have not time for this feature now.
Yes, @asterixcapri is  right. I implement simple solution for fix console bug, but my PR add some changes in public interfaces.
And  also I have strange activity after this, like [this](https://github.com/Block8/PHPCI/issues/509#issuecomment-51688055).

This solution need more testing, I can do it, but in March.

dancryer:
Thanks @nonlux - I rebased your changes and applied them. :)

nonlux:
:ok_hand:

PR #566, opened 2014-08-21
Gitlab CI merge request service

mavimo:
This is a first implementation of exposed entrypoint to build gitlab ci service to integrate PHPCI as CI server.

Actually I have JSON conversion into Gitlab class, maybe will be better have a more generic helper to allow all type of build to be converted on JSON. Before I implemented it on controller and after moved on ```PHPCI\Model\Build``` but I'm not sure is a good place to store this conversion. Maybe a new helper class? Any suggestion?

I think is required a token to allow information reading without expose project information as public.

Test are required, please do not merge, just use this MR as support request :smile:.

Next steps:

 - [ ] define better solution to convert build to JSON
 - [x] introduce a validation token to allow non public project to be read
 - [ ] add tests
 - [x] implement phpci_ci_service similar to [gitlab_ci_service](https://github.com/gitlabhq/gitlabhq/blob/master/app%2Fmodels%2Fproject_services%2Fgitlab_ci_service.rb)

### DIFF
--- a/PHPCI/Controller/BuildStatusController.php
+++ b/PHPCI/Controller/BuildStatusController.php
@@ -149,7 +149,7 @@ class BuildStatusController extends \PHPCI\Controller
 
         if (!$build) {
             $this->response->setResponseCode(404);
-        } elseif (!$build->getProject()->getAllowPublicStatus()) {
+        } elseif (!$build->getProject()->allowAccess($this->request->getParam('auth_token'))) {
             $this->response->setResponseCode(401);
         } elseif ($build) {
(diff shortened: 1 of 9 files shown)


tvbeek:
@mavimo Did you have a way to finish this feature? Maybe it is an option to create also a pull request to gitlab if this is finished.

mavimo:
@tvbeek I changed approach since gitlab PR are not accepted since there are a lot of push on it.. I will create a controller that have the same fingerprint of Gitlab CI, and use the same service definition. I will send it as contrib ASAP.

PR #576, opened 2014-09-01
[FEATURE] Add option "allowedBranches" and variables in filename to PackageBuild plugin

ArminVieweg:
We had the problem that each commit in any branch creates a package. This creates a lot of waste we don't need. We just want packages of the master and develop branch. So with this option you are able to restrict this plugin to the given branches. The option expects an regular expression like this: 

```
  PackageBuild:
    directory: "/.../builds/%project.title%/%build.branch%/"
    filename: "build-%build.id%"
    allowedBranches: "^master$|^develop$"
    format: tar
```

Furthermore now the filename may also contain variables, which become replaced.

### DIFF
--- a/PHPCI/Plugin/PackageBuild.php
+++ b/PHPCI/Plugin/PackageBuild.php
@@ -33,7 +33,7 @@ class PackageBuild implements \PHPCI\Plugin
         $this->directory       = isset($options['directory']) ? $options['directory'] : $path;
         $this->filename        = isset($options['filename']) ? $options['filename'] : 'build';
         $this->format          = isset($options['format']) ?  $options['format'] : 'zip';
-		$this->allowedBranches = isset($options['allowedBranches']) ?  $options['allowedBranches'] : NULL;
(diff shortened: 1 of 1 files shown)


mavimo:
@ArminVieweg have you time to fix the reported errors?

ArminVieweg:
Maybe next month

mavimo:
refs #919 

@ArminVieweg any news? :)

mavimo:
@ArminVieweg ping

ArminVieweg:
@mavimo pong

Sorry I am very busy currently. But in the meantime I thought about this feature request and wouldn't it be much more flexible to add a global option which limits execution to branches? Like the ignore option.

ArminVieweg:
Btw. I've forked your project, I don't know if github tells you that, when importing it in Bitbucket, probably not... but here is the link: https://bitbucket.org/ArminVieweg/phpci

Maybe some ideas you want to adopt during further development :)

PR #624, opened 2014-10-17
TapParser must capture skipped tests.

wandersonwhcr:
This PR must be applied before PR #611, because we need to consider skipped tests too. I don't know if this is the best solution; other ideas are welcome.


tjantika:
This support for skipped tests may not be enough. I have a test class where one test method is marked skipped and reveral other test methods depend on that, so they get skipped too. I also have one test method in that class that is not skipped so the tap output looks like this:

TAP version 13
ok 1 - # SKIP Skipped due to x
ok 1 - # SKIP This test depends on x to pass.
ok 1 - # SKIP This test depends on x to pass.
ok 1 - # SKIP This test depends on x to pass.
ok 1 - # SKIP This test depends on x to pass.
ok 1 - # SKIP This test depends on x to pass.
ok 1 - # SKIP This test depends on x to pass.
ok 2 - MyTestClass::myTestMethod
ok 2 - # SKIP This test depends on y to pass.
1..2

so basicly the number of tests run ($totalTests in TapParser::parse()) does not match at all with the number of the lines in the output.

Issue #571, opened 2014-08-24
Invalid TAP string, number of tests does not match specified test count.


tobiastom:
I've added this login section to the `phpunit.xml` file:

```
<logging>
    <log type="coverage-html" target="Build/Coverage/" charset="UTF-8" highlight="false" lowUpperBound="35" highLowerBound="70"/>
    <log type="coverage-clover" target="Build/Coverage/index.xml"/>
    <log type="coverage-text" target="Build/Coverage/index.txt" showUncoveredFiles="false"/>
    <log type="tap" target="Build/Coverage/index.tap"/>
</logging>
```

The result seems to be:

```
TAP version 13
ok 1 - …
ok 2 - …
…
ok 45 - …
ok 46 - …
1..46
Generating code coverage report in Clover XML format ... done
Generating code coverage report in HTML format ... done
```

The error I get is this:

```
EXCEPTION: Invalid TAP string, number of tests does not match specified test count.
PLUGIN STATUS: FAILED
```

Caffe1neAdd1ct:
Hi Guys,

Just started using PHPCI, absolutly brilliant project! Had this up and running on my local machine in under 5 mintues :+1: 

Just came across this issue:

```
RUNNING PLUGIN: php_unit
TAP version 13
TAP version 13
ok 1 - Billing_Model_CountryServiceTest::testThisFunctiRUNNING PLUGIN: php_unit
TAP version 13
TAP version 13
ok 1 - Billing_Model_CountryServiceTest::testThisFunction
ok 1 - Billing_Model_CountryServiceTest::testThisFunction
ok 2 - Billing_Model_CountryTest::testGetId
ok 2 - Billing_Model_CountryTest::testGetId
ok 3 - Billing_Model_CountryTest::testGetName
ok 3 - Billing_Model_CountryTest::testGetName
ok 4 - Billing_Model_CountryTest::testGetCode
ok 4 - Billing_Model_CountryTest::testGetCode
ok 5 - Billing_Model_CountryTest::testGetVatChargableFlag
ok 5 - Billing_Model_CountryTest::testGetVatChargableFlagon
ok 1 - Billing_Model_CountryServiceTest::testThisFunction
ok 2 - Billing_Model_CountryTest::testGetId
ok 2 - Billing_Model_CountryTest::testGetId
ok 3 - Billing_Model_CountryTest::testGetName
ok 3 - Billing_Model_CountryTest::testGetName
[...]

PR #676, opened 2014-12-04
Issue #571 Ignoring Duplicated Output

wandersonwhcr:


dancryer:
@wandersonwhcr This seems like an odd thing to have to ignore. Any idea why the duplicated output is happening in the first place?

dancryer:
@wandersonwhcr Are you still experiencing this issue or are we safe to close this PR?

Issue #697, opened 2014-12-10
Hooks on Merge Requests (GitLab)


lukzgois:
I have configured my gitlab repository to create a build when a new merge request is open.
When I hit the test hook button it works normaly, but when a merge request is created it doesn't.

Any idea about what is happening?

PR #699, opened 2014-12-11
Added support for gitlab merge requests

lukzgois:
Added function to verify if a gitlab merge request is sent.
If yes, then will create a build for the source branch of the request.

Refs to #697


lukzgois:
Why it was marked as a failure?

mavimo:
@lukzgois somo PHPCS issue, but is already merged in e4bf5ee

PR #723, opened 2014-12-25
Added ext-mcrypt as requirement in composer.json file.

mavimo:
The PHP Mcrypt extension is required to allow PHPCI execution. Added this dependency in the ```composer.json``` file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
 
     "require": {
         "php": ">=5.3.8",
+        "ext-mcrypt": "*",
         "ext-pdo": "*",
         "ext-pdo_mysql": "*",
         "block8/b8framework": "~1.0",


dancryer:
I don't believe PHPCI uses mcrypt as standard @mavimo. If it does, we need to remove that dependency rather than adding it to composer.json. :)

denji:
* http://news.php.net/php.internals/91853
* https://wiki.php.net/rfc/libsodium
* https://paragonie.com/blog/2015/09/state-cryptography-in-php
* https://paragonie.com/book/pecl-libsodium/read/00-intro.md

By the way you need to think about how to get rid of dependence mcrypt.
mcrypt does not develop more than 10 years and will soon be deleted.

PR #725, opened 2015-01-04
Resetting the status of a duplicated build

dansoap:
When using SQL strict mode, mysql would complain that no default value of the status column is set. Setting the status to 0 before duplicating fixes this.


mavimo:
:+1:

dansoap:
Any chance of having this merged soon? I'm still running into this error ;-)

PR #730, opened 2015-01-08
Email plugin: use "default_mailto_address" as a fallback only.

Adirelle:
I'm not sure what the intended behavior of the "default_mailto_address" parameter was, but right now it works exactly like "addresses".

With this change, the "default_mailto_address" is only used is the recipient list is empty, e.g. with no commiter email nor addresses.

### DIFF
--- a/PHPCI/Plugin/Email.php
+++ b/PHPCI/Plugin/Email.php
@@ -178,10 +178,10 @@ class Email implements \PHPCI\Plugin
             }
         }
 
-        if (isset($this->options['default_mailto_address'])) {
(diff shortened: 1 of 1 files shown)


Adirelle:
I noticed there is a "default mailto address" in PHPCI global configuration. Is it related to the phpci.yml address in some way ?

dancryer:
Hi @Adirelle - Please could you rebase this change on top of the current master branch?

Adirelle:
Yes, I can but I'm not sure about the purpose of the default address. Is it related to the "default mailto address" entry in the settings ?

PR #731, opened 2015-01-08
Email plugin: remove duplicates from the list of recipients.

Adirelle:
Simply calls "array_unique" on the recipient list. This avoids sending duplicate emails if the committer is also in "addresses".


mavimo:
Add test, please.

PR #733, opened 2015-01-08
French translation updates.

Adirelle:


PHPCI/Languages/lang.fr.php
```diff
@@ -67,19 +67,19 @@
...
-    'no_builds_yet' => 'Aucun build pour le moment!',
-    'x_of_x_failed' => '%d parmis les derniers %d builds ont échoué.',
-    'x_of_x_failed_short' => '%d / %d ont échoué.',
-    'last_successful_build' => ' Le dernier build qui a réussi est %s.',
-    'never_built_successfully' => ' Aucun build n\'a été exécuté avec succès sur ce projet.',
-    'all_builds_passed' => 'Les derniers %d builds ont réussis.',
-    'all_builds_passed_short' => '%d / %d ont réussis.',
-    'last_failed_build' => ' Le dernier build en échec est %s.',
-    'never_failed_build' => ' Ce projet n\'a jamais eu un build en échec.',
+    'no_builds_yet' => 'Aucun build pour le moment !',
```
mavimo: Space between text and punctuation is not good.

Adirelle: In French, you have to add a (unbreakable) space before "double punctuation" marks: colon, semi-colon, interrogation and exclamation marks (see https://en.wikipedia.org/wiki/Punctuation#Other_languages). However, I'm not sure how to represent this in this case. The UTF-8 character ( ) ? The HTML code (&#160;) ? The entity (&nbsp;) ? Other ?

mavimo: I will prefer ```&nbsp;``` (No-break space) that specify the brevius word and punctation shuldn't be splitted.

mavimo:
Some space still need to be replaced.

Adirelle:
I think these spaces are inside strings used only in the console, where `&nbsp;` would be quite ugly.

mavimo:
@Adirelle you're right, thx!

PR #734, opened 2015-01-09
Full front-end date localization using <time> tags and moment.js.

### LINKED ISSUE #732: Date localization
Even though the front-end uses the right language, most dates are still in English format (field ordering, AM/PM).

I would like to submit a request, but I am not sure how to fix this:
* for PHP-rendered dates, could we use a PHP module, like the intl module ? (see http://php.net/manual/fr/class.intldateformatter.php) or use built-in code ?
* for the moment.js-rendered dates, I guess `moment.locale(currentLang);` should be added somewhere in the page initialization.

Any answer ?

Adirelle:
Should fix #732.

I had to update the daterangepicker with a version that does not embed an old version of moment.js.

I added a PHPCI\Helper\Lang::formatDateTime method to create the <time> tags but I am not sure it belongs to that class. This helper creates the tag with a default text and the required attributes so it can get reformated with moment.js on DOMReady.

I think there is still an issue with locale and AJAX (i.e. when the project view is refreshed) but this is not specifically related to dates.

### DIFF
--- a/PHPCI/Helper/Lang.php
+++ b/PHPCI/Helper/Lang.php
@@ -168,4 +168,20 @@ class Lang
             }
         }
     }
+
+    /**
+     * Create a time tag for localization.
+     * @param \DateTime $dateTime The dateTime to represent.
+     * @parma string $format The moment.js format to use.
+     * @return string The formatted tag.
+     */
+    public static function formatDateTime(\DateTime $dateTime, $format = 'lll')
+    {
+        return sprintf(
+            '<time datetime="%s" data-format="%s">%s</time>',
+            $dateTime->format(\DateTime::ISO8601),
+            $format,
+            $dateTime->format(\DateTime::RFC2822)
+        );
+    }
 }
--- a/PHPCI/Helper/Lang.php
+++ b/PHPCI/Helper/Lang.php
@@ -172,7 +172,7 @@ class Lang
     /**
      * Create a time tag for localization.
      * @param \DateTime $dateTime The dateTime to represent.
-     * @parma string $format The moment.js format to use.
+     * @param string $format The moment.js format to use. See http://momentjs.com/docs/#/displaying/format/
      * @return string The formatted tag.
      */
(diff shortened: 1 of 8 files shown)


PHPCI/Helper/Lang.php
```diff
@@ -168,4 +168,20 @@ protected static function loadAvailableLanguages()
             }
         }
     }
+
+    /**
+     * Create a time tag for localization.
+     * @param \DateTime $dateTime The dateTime to represent.
+     * @parma string $format The moment.js format to use.
```
mavimo: Typo on ```@param```

PHPCI/Helper/Lang.php
```diff
@@ -168,4 +168,20 @@ protected static function loadAvailableLanguages()
             }
         }
     }
+
+    /**
+     * Create a time tag for localization.
```
mavimo: Empty line between function description, params and from params and return value.

public/assets/js/phpci.js
```diff
@@ -14,6 +20,9 @@ var PHPCI = {
             }
         });
 
+        // Setup the date locale
+        moment.locale(PHPCI_LANGUAGE);
```
mavimo: Move this init before document ready (don't change the behavior, but make the code more understandable).

mavimo:
Please add test to check the tag creation function.

PHPCI/Helper/Lang.php
Adirelle: The other methods of the class don't match this format. Do I proceed anyway ?

mavimo: https://github.com/Block8/PHPCI/wiki/Contributing-to-PHPCI#coding-standards

> ... If you're modifying a file as part of your change which does not comply with the above, please make the necessary changes to bring it into compliance prior to submitting the pull request.

[PhpDocumentor](http://www.phpdoc.org/docs/latest/getting-started/your-first-set-of-documentation.html#what-does-a-docblock-look-like) references.

Yes, please :smile:

PR #738, opened 2015-01-12
Convert tapString to UTF-8

mattleff:
My php_unit suite includes some non-UTF8 characters in the datasets that are passed to the tests. This was causing the `json_encode()` in `PHPCI\Model\Build::storeMeta()` to fail with `JSON_ERROR_UTF8`. I'm not sure if you have an abstraction for character type conversion or if there is a better place to convert this, but this is what fixed my issue. :smile:

### DIFF
--- a/PHPCI/Plugin/PhpUnit.php
+++ b/PHPCI/Plugin/PhpUnit.php
@@ -162,7 +162,7 @@ class PhpUnit implements PHPCI\Plugin, PHPCI\ZeroConfigPlugin
         }
 
         $tapString = $this->phpci->getLastOutput();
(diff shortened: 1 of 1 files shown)


mavimo:
# PHP Code Sniffer
File | Line | Message
-----|------|----------
PHPCI/Plugin/PhpUnit.php | 165 | Space after opening parenthesis of function call prohibited
PHPCI/Plugin/PhpUnit.php | 165 | Expected 0 spaces before closing bracket; 1 found

mattleff:
@mavimo I've pushed a commit with the CS fixes.

PR #742, opened 2015-01-14
Fix slack message send

ViGo5190:
Fix slack problem. Now it send message correctly

Contribution Type: bug fix 
Primary Area: plugins

Description of change: create $message instead of using $client

Description of solution: Fix slack problem. Now it send message correctly

Issue:#747


ViGo5190:
#747

REBELinBLUE:
Hmm this is odd, what error are you getting, I use this every day without issue.

$client->send() isn't a real method of \Maknz\Slack\Client, but is caught by __call, which then calls createMessage() on $client and calls the send method so it should be doing the same thing

REBELinBLUE:
I see the problem, when I wrote it maknz/slack was at version 0.2.2, the latest version is 1.1.0 and has indeed broken the plugin

PR #751, opened 2015-01-16
Add Jasmine & Pho testing support

AustP:
Contribution Type: new plugins
Primary Area: plugins

Adding a link for a new [plugin](https://github.com/austp/phpci-pho-plugin) to support [pho](https://github.com/danielstjules/pho)
Adding a link for a new [plugin](https://github.com/austp/phpci-jasmine-node-plugin) to support [Jasmine](https://jasmine.github.io) via [jasmine-node](https://github.com/mhevery/jasmine-node)


mavimo:
I think is better to split this in two different PR.

AustP:
Ok I'll split it now.

PR #754, opened 2015-01-21
Update HipChat plugin to support self-hosted servers

bertramtruong:
This PR will allow users to specify a "target" for hipchat_notifier so that they can send notifications to a self-hosted [HipChat Server](https://www.hipchat.com/server) instance.


mavimo:
Please add test to check the correct behavior when the target is set.

Also add a check connection error if hipchat config is wrong.

dancryer:
Hey @bertramtruong - Please could you rebase this change on top of the current master branch? 

Otherwise, :+1: from me.

PHPCI/Plugin/HipchatNotify.php
```diff
@@ -52,6 +53,12 @@ public function __construct(Builder $phpci, Build $build, array $options = array
                 $this->message = Lang::get('x_built_at_x');
             }
 
+            if (isset($options['target'])) {
```
mavimo: This should be done by ternary operator:
```php
 $this->target = isset($options['target']) ? $options['target'] : "https://api.hipchat.com";
```

mavimo:
@bertramtruong any chance to be update / rebased?

PR #749, opened 2015-01-15
Fix issue Phing targets are not executed in the build dir

corretge:
See comments in the issues #746 and #748 and commits 56f04d9 and ed13ec5.

Other commits without linked official issue has been attached to the pull request. Maybe could be interesting for the project:

* #746 fix Field 'status' doesn't have a default value when click "Rebuild Now"
* #748 fix Phing targets are not executed in the build dir
* Fix issue build_settings ignored in createWorkingCopy.
* Add a new option in the **build** section: prefer_rsync that exclude the ignored folders in the copy when the project is hosted in **Local Path**.
* Flush stdout while executing a step, don't wait at the end of the execution.
* Improve to log more colours from console stdout than red and green.
* Allow to override phpci.yml in a concrete build.
* Fix phpunit bootstrap that crash if not exists a config file.

Issue #143, opened 2013-08-30
Add support for ApiGen


gregrobson:
Add support for ApiGen for creating documentation as it can produce a "checkstyle" report for undocumented functions/parameters (as well as generating some pretty nice looking documentation of your code).

http://apigen.org/
https://github.com/apigen/apigen

I'm tempted to give it a go myself but I just spent a couple of days getting it working on IIS/Windows (don't ask!). I might see if I can put together some pull requests for that first!

mavimo:
the signature for CLI (Apigen 4.0.0) is the same for ```phar``` and ```bin```, eg:

    php apigen.phar generate --config /tmp/apigen.neon -d /tmp/dest -s /tmp/src
    php vendor/bin/apigen generate --config /tmp/apigen.neon -d /tmp/dest -s /tmp/src

wandersonwhcr:
Do you know why ```dirname``` the ```BUILD_PATH```?

https://github.com/Block8/PHPCI/blob/master/PHPCI/Helper/BaseCommandExecutor.php#L95

If I understood, ```BUILD_PATH``` can be ```APPLICATION_PATH/PHPCI/build/123```. With ```dirname```, ```proc_open``` is using as cwd a parent directory, in this case ```APPLICATION_PATH/PHPCI/build```.

Checking other plugins, like ```PHPCI\Plugin\PhpCodeSniffer```, there's a binary parameter to change the "working directory", but ApiGen there's not.

Any suggestions?

wandersonwhcr:
OK, I debugged and `$this->basePath` is **always** `NULL`.

So, `dirname(NULL)` is equals to an empty string and if `proc_open` receives an empty string as cwd, it will use the current working directory.

The default cwd to crontab is the HOME user directory. So, if you configure the `php console phpci:run-builds` inside the root crontab without a `cd`: Houston, do we have a problem?

https://github.com/Block8/PHPCI/blob/master/PHPCI/Helper/BaseCommandExecutor.php#L95

I'll create a patch.

Adirelle:
I've noticed some discrepancies amongst plugins with regard to the current directory and the build path. This may explain why some of them are failing. IMHO, the whole build process should be executed in the build path by default.

wandersonwhcr:
I'm sorry, typo: `$this->buildPath` is always `NULL`.

@Adirelle I agree with you. I think we must correctly configure the `buildPath` and stack cwd before the plugin execution.

PR #756, opened 2015-01-22
ApiGen Plugin

wandersonwhcr:
Olá a todos! :smiley:

# ApiGen Plugin `api_gen`

I develop a super simple ApiGen plugin according to #143. You'll only need to add an `apigen.neon` configuration in the root of your project.

More information in https://github.com/apigen/ApiGen.

## Notices

1. I don't know if my `composer.lock` change is right. Future problems with merging? Probably;
2. Plugin supports only `apigen >= 4.*`. Composer `suggest` option don't support versions;
3. Stage "success" and plugins without options cause PHP Fatal Error. 8ca0a04 and 86157cc. Fixed;
4. I called a `chdir` before and after binary execution;
5. I can't use vfs to create a project structure because `apigen` is called with `proc_open`; and
6. My PR will only pass in PHPCI if PR #749 55a2b79 is applied; thanks @corretge.

## Usage

Add to your `phpci.yml`:

```
success:
    api_gen:
```

### DIFF
--- a/PHPCI/Plugin/ApiGen.php
+++ b/PHPCI/Plugin/ApiGen.php
@@ -1,0 +1,60 @@
+<?php
+namespace PHPCI\Plugin;
+
+use PHPCI\Builder;
+use PHPCI\Model\Build;
+
+class ApiGen implements \PHPCI\Plugin
+{
+    /**
+     * PHPCI
+     * @var Builder
+     */
+    protected $phpci;
+
+    /**
+     * Build
+     * @var Build
+     */
+    protected $build;
+
+    /**
+     * Standard Constructor
+     *
+     * @param Builder $phpci
+     * @param Build   $build
+     * @param array   $options
+     */
+    public function __construct(Builder $phpci, Build $build, array $options = array())
+    {
+        // Basic
+        $this->phpci = $phpci;
+        $this->build = $build;
+    }
+
+    /**
+     * Returns PHPCI
+     *
+     * @return PHPCI
+     */
+    public function getPHPCI()
+    {
+        return $this->phpci;
+    }
+
+    /**
+     * Returns Build
+     *
+     * @return Build
+     */
+    public function getBuild()
+    {
+        return $this->build;
+    }
+
+    // Execution
+    public function execute()
+    {
+        return true;
+    }
+}
--- a/PHPCI/Plugin/ApiGen.php
+++ b/PHPCI/Plugin/ApiGen.php
@@ -4,6 +4,9 @@ namespace PHPCI\Plugin;
 use PHPCI\Builder;
 use PHPCI\Model\Build;
 
+/**
+ * PHPCI ApiGen Plugin
+ */
 class ApiGen implements \PHPCI\Plugin
 {
     /**
@@ -23,9 +26,8 @@ class ApiGen implements \PHPCI\Plugin
      *
      * @param Builder $phpci
      * @param Build   $build
-     * @param array   $options
      */
-    public function __construct(Builder $phpci, Build $build, array $options = array())
+    public function __construct(Builder $phpci, Build $build)
     {
         // Basic
         $this->phpci = $phpci;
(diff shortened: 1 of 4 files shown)


wandersonwhcr:
PHPMD problem. I'll check soon.

wandersonwhcr:
gotcha!

wandersonwhcr:
## Notices

6.2 or PR #758

dancryer:
@wandersonwhcr This looks great, thanks :) 

I'm intrigued what the changes to the Builder class are for?

wandersonwhcr:
@dancryer I created tests about it: `Tests/PHPCI/Build/BuilderTest.php`.

If we configure an empty stage, PHP generate a fatal error, because it is expecting an array and receives a null, exactly here:

https://github.com/Block8/PHPCI/blob/master/PHPCI/Plugin/Util/Executor.php#L57

If we configure an empty `test` and `allow_failures`, PHP automagically configures `$options` variable to array at line 53.

Example:

```yaml
success:
```

or

```yaml
success:
    api_gen:
```

Thinking about it now, I just need to create another test, combining Builder and PluginExecutor. Sorry about that.

dancryer:
Hi @wandersonwhcr - Please could you rebase this change on top of the current master branch? :)

Otherwise, :+1: from me.

Tests/PHPCI/Build/BuilderTest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+namespace PHPCI\Build\Tests;
+
+use Symfony\Component\Yaml\Parser as YamlParser;
+
+class BuilderTest extends \PHPUnit_Framework_TestCase
+{
+    public function testExecute_EmptyStageConvertNullToArray()
+    {
+        $builder = $this
```
mavimo: Move builder in a class variable and configure in ```setUp()```.

Tests/PHPCI/Build/BuilderTest.php
```diff
@@ -0,0 +1,37 @@
...
+            ->getMockBuilder('PHPCI\Builder')
+            ->setMethods(null)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $config = (new YamlParser())->parse("stage:\n");
```
mavimo: This should be only an array?

Tests/PHPCI/Build/BuilderTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+    public function testExecute_PluginWithNoParamsConvertNullToArray()
+    {
+        $builder = $this
+            ->getMockBuilder('PHPCI\Builder')
+            ->setMethods(null)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $config = (new YamlParser())->parse("stage:\n    plugin:\n");
```
mavimo: This should be only an array?

Tests/PHPCI/Build/BuilderTest.php
wandersonwhcr: Ack. Commit 5c2a4e77.

Tests/PHPCI/Build/BuilderTest.php
wandersonwhcr: Verifying the current source code, the `PHPCI\Plugin\Util\Executor::buildPlugin` is not expecting the second parameter as array, because of dd58dd68. I don't know why a plugin configuration must not always be an array, but, I'll remove my changes about it.

wandersonwhcr:
I refactored my code, cleaning some dirty commits and removing my `YamlParser` patch, because dd58dd68 removed the expected array value at `PHPCI\Plugin\Util\Factory::buildPlugin` method.

But now I don't know why my source code don't pass in PHPCI. I cloned your source code and checked out `1.6.0` tag and I installed it. I added a project in PHPCI, pointing to my local path. When I build locally, all works fine.

Can you send me the errors from fd98b20 commit?

mavimo:
@wandersonwhcr There is a PHPUnit test that fail:

Failure: PHPCI\Plugin\Tests\ApiGenTest::testSuccess
Failed asserting that false is true.

PR #644, opened 2014-11-11
E-mail add support for sending via sendmail

mavimo:
This PR:
 * Remove dependency from swiftmail for the email plugin.
 * Introduce a mail template also for successful build.
 * Code cleanup removing unrequired dependencies.
 * CTest refactoring to remove dependencies from Swiftmail and having a isolated component to test (mail options, status code, subject, body, ..).

refs #598 cc/ @dancryer @tvbeek

PR #758, opened 2015-01-24
Fix undefined variable '$config' in /Tests/bootstrap.php

thaJeztah:
When no config-file was found, no instance of Config was created as well, resulting in:

PHP Notice:  Undefined variable: config in /Tests/bootstrap.php on line 41

This changes /Tests/bootstrap.php to always create an instance of b8\Config(), even if no file was found.

Signed-off-by: Sebastiaan van Stijn <github@gone.nl>

### DIFF
--- a/Tests/bootstrap.php
+++ b/Tests/bootstrap.php
@@ -31,8 +31,9 @@ if (!file_exists($configFile)) {
     }
 }
 
+$config = new b8\Config($conf);
+
 if (file_exists($configFile)) {
-    $config = new b8\Config($conf);
     $config->loadYaml($configFile);
 }
 


wandersonwhcr:
DUP @corretge PR #749 55a2b79

thaJeztah:
Thanks for finding that, @wandersonwhcr 

Wow, that's a *huge* PR; would it be better to break that up in to several smaller ones?

wandersonwhcr:
@thaJeztah sure, I agree with you :)

wandersonwhcr:
But I think these PR's can cause some conflicts if applied both.

thaJeztah:
Yup; if this one gets merged the other PR needs a rebase, the other way around, this PR just becomes obsolete.

Overall, the other PR looks to have many not-directly related fixes, so I would be in favor of breaking it up to smaller changes, but that's something the maintainers here should decide.

We'll see what happens! Thanks for the heads-up, I don't think I would've found it otherwise :)

wandersonwhcr:
No problem. I just need this patch to my ApiGen plugin works. So, thank you! :smiley:

mavimo:
fixed also on https://github.com/Block8/PHPCI/pull/644. I think we need a check on the ready-to-merge request from maintainer in order to reduce the number of duplicate issue/PR.

cc/ @dancryer @tvbeek

PR #761, opened 2015-01-26
Add Jasmine testing support

AustP:
Contribution Type: new plugins
Primary Area: plugins

Adding a link for a new [plugin](https://github.com/austp/phpci-jasmine-node-plugin) to support [Jasmine](https://jasmine.github.io) via [jasmine-node](https://github.com/mhevery/jasmine-node)

Reference #751

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,6 +65,7 @@
         "behat/behat": "Behat BDD Testing",
         "hipchat/hipchat-php": "Hipchat integration",
         "phptal/phptal": "PHPTAL templating engine",
-        "maknz/slack": "Slack integration"
+        "maknz/slack": "Slack integration",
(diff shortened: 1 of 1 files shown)


dancryer:
@AustP - Happy to merge - Please can you write up some documentation for your Jasmine plugin and post it as a comment here?

dancryer:
Having realised what this PR does, I'm not sure it's something we should merge just now. 

We tend to only include in "suggest" the things that are required by built-in PHPCI plugins. I think suggesting external plugins might be confusing. 

Any thoughts @Block8/phpci-contributors?

AustP:
Would you consider including these plugins as part of the built-in plugins?

dancryer:
@AustP Yes, we would accept this as a built in plugin. However, I don't want to add "suggested" dependencies that, even if installed, PHPCI by default cannot use.

Issue #762, opened 2015-01-27
Ability to add project to archive


zviryatko:
Hi,
my team has a new project nearly every two week, so we have many-many project and projects list in main screen is big, so is there any such functionality in the plans for the future?

I think that it is better to sort projects since the last build and hide archived project from main projects list (but show it in right side if project has new commit, for gitlab hooks).

I will try implement this feature, so write your suggestions.

tvbeek:
I like the idea.
I think it is good to have the option to sort on last date (but not to replace the sort on name)
And an option to archive the project so it is hidden from the main page. (It need to have a filter or separated page with archived projects to set it back)

maschmann:
:+1:

PR #771, opened 2015-01-31
Archive projects

zviryatko:
As I suggested in #762 it allow move project to archive.


zviryatko:
How can I see more details? I don't have account on http://phpci.block8.net. Think this information needs to be added to wiki pages.

PHPCI/Application.php
```diff
@@ -133,7 +133,7 @@ protected function setLayoutVariables(View &$layout)
     {
         /** @var \PHPCI\Store\ProjectStore $projectStore */
         $projectStore = b8\Store\Factory::getStore('Project');
-        $layout->projects = $projectStore->getAll();
+        $layout->projects = $projectStore->getWhere(array('archive' => (int)isset($_GET['archive'])), 50, 0, array(), array('title' => 'ASC'));
```
dancryer: PHPCS Error: Line exceeds 120 characters; contains 143 characters

PHPCI/Controller/HomeController.php
```diff
@@ -48,7 +48,7 @@ public function index()
     {
         $this->layout->title = Lang::get('dashboard');
 
-        $projects = $this->projectStore->getWhere(array(), 50, 0, array(), array('title' => 'ASC'));
+        $projects = $this->projectStore->getWhere(array('archive' => (int)isset($_GET['archive'])), 50, 0, array(), array('title' => 'ASC'));
```
dancryer: PHPCS Error: Line exceeds 120 characters; contains 141 characters

PHPCI/Model/Base/ProjectBase.php
```diff
@@ -148,6 +151,9 @@ class ProjectBase extends Model
             'type' => 'int',
             'length' => 11,
         ),
+        'archive' => array(
+            'type' => 'boolean',
```
dancryer: Did you modify this file by hand, or did you use ./console db:generate to update it? I'm intrigued that it used a "type" value of boolean.

PHPCI/Migrations/20150131075425_archive_project.php
```diff
@@ -0,0 +1,26 @@
...
+
+class ArchiveProject extends AbstractMigration
+{
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $project = $this->table('project');
+        $project->addColumn('archived', 'boolean');
```
dancryer: Everywhere else, the column is referred to as `archive` - here you've used `archived`.

As it happens, I actually prefer "archived".

dancryer:
Hi @zviryatko - A few fixes to make, but otherwise this looks good. :+1:

PHPCI/Model/Base/ProjectBase.php
zviryatko: I didn't know about db:generate, I've added it by hand. Where I can read more about it?

dancryer: @zviryatko The best practice is:

1. Make a migration.
2. Run it.
3. Run `./console db:generate`

That'll update the database, and then update your models to reflect the changes to the database. 

The *Base.php files should never be edited by hand.

PHPCI/Migrations/20150131075425_archive_project.php
zviryatko: Firstly I've added "archive", than search in popular systems how to call column and decided to call it archive, but ok, I rename it to archived.

PHPCI/Model/Base/ProjectBase.php
zviryatko: Can you add it to documentation for contributors?

zviryatko:
Hi, I've fixed all previous things and don't know what is wrong now, can you post failure log?

mavimo:
## PHP Code Sniffer

File | Line | Message
-----|------|--------
PHPCI/Application.php | 137 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Application.php | 138 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Application.php | 139 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Application.php | 140 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Application.php | 141 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Controller/HomeController.php | 52 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Controller/HomeController.php | 53 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Controller/HomeController.php | 54 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Controller/HomeController.php | 55 | Multi-line function call not indented correctly; expected 12 spaces but found 10
PHPCI/Controller/HomeController.php | 56 | Multi-line function call not indented correctly; expected 12 spaces but found 10

PHPCI/Application.php
```diff
@@ -133,7 +133,13 @@ protected function setLayoutVariables(View &$layout)
     {
         /** @var \PHPCI\Store\ProjectStore $projectStore */
         $projectStore = b8\Store\Factory::getStore('Project');
-        $layout->projects = $projectStore->getAll();
+        $layout->projects = $projectStore->getWhere(
+          array('archived' => (int) isset($_GET['archived'])),
```
mavimo: This mean that if ```archived``` is ```false``` or ```0``` we always display also archived projects because is set, eg:

    http://phpci/controller/action?archived=false

have ``isset($_GET['archived']);``` as true and can be confusing, maybe we need to check for isset and it's value.

PHPCI/Languages/lang.en.php
```diff
@@ -113,6 +113,7 @@
                                 (if you cannot add a phpci.yml file in the project repository)',
     'default_branch' => 'Default branch name',
     'allow_public_status' => 'Enable public status page and image for this project?',
+    'archived' => 'Archived',
```
mavimo: There is a new language string here, maybe is a good idea to add it also to other languages, so translators can see it and update it if required, @dancryer what do you think?

zviryatko: I can't understand why you don't use .po/.mo file types for translations, benefits in .pot file that contains all original strings and any LANG.po files can be updated from .pot file.

PHPCI/Application.php
zviryatko: I think it make sense when `$_GET[archive]` has multiple values, it's good for `status` parameter, like: unpublished, published, archived or any other, in this case it's logical.

PHPCI/Languages/lang.en.php
mavimo: I agree about better language management, but actually is not implemented, maybe you can open a different PR about it..

zviryatko: Ok, I'll try.
So what about this PR? =)

mavimo:
@zviryatko can you rebase please.

dancryer:
Hi @zviryatko - I really like this addition, but I can't get the PR to apply cleanly. Please can you give this a rebase on top of the current master branch and see where we get to?

Thanks!

zviryatko:
I'm not familiar with advanced git workflow with branches and forks, so not immediately understand what must to do. Now everything is fine. Check it, please.

zviryatko:
Interesting, why in layout.phtml link to archive moved after projects https://github.com/Block8/PHPCI/commit/1466ad06ef708cbab2b53112fc59e8c1d70c2e33#diff-4c3dcb804d8e696bfa0db20c8e5c6860R245, but in my last commit it right after the dashboard link https://github.com/zviryatko/PHPCI/commit/161a8a5ecf6fe8a358c0a37f16135581a71c5672#diff-4c3dcb804d8e696bfa0db20c8e5c6860R150

dancryer:
@zviryatko Because the archive is, by definition, less important than projects that are not archived. I didn't want it to have such precedence in the menu. :)

zviryatko:
I did not know it was you did, I thought something went wrong =)

PR #773, opened 2015-02-01
Move legends to below chart to avoid word wrapping issues

leewillis77:
The standard Lines of Code, and Quality Trend charts have the legend to the write. On my display at least, this causes word wrapping issues that render the legends pretty much useless.

This change moves the legends to below the chart which improves things considerably. 

Before the change:
https://www.dropbox.com/s/55yrz602a3h1uba/Screenshot%202015-02-01%2015.20.07.png?dl=0

After the change:
https://www.dropbox.com/s/1qiw8nakpnj2c48/Screenshot%202015-02-01%2015.21.58.png?dl=0


dancryer:
Thanks @leewillis77 - This has been merged.

PR #774, opened 2015-02-01
Fixes and improvements to german translation

pulponair:
Fixes 2 typos.

Rephrases the "build_now" from "Aktueller Build"  to "Jetzt bauen"

The old variant "Akueller Build" is not correct since it reads like "Recent build" while this button actually triggers a build. 

"Jetzt bauen" fits way better and is what e.g. the jenkins "build now" button is translated to ;).


dancryer:
Thanks @pulponair - This has been merged.

Issue #708, opened 2014-12-16
about the dashboard projects status


linuxjuggler:
I think that it would be nice if the got the color based on the last build status I mean if the last build was a success it should be green , even if he had for example 3 builds failed out of 5 .

the visuality here should be based on the last build not the whole project status.

danaketh:
I think simple `Project health: N%` would be enough.

leewillis77:
Independent to this issue I mocked this up earlier today. I like the comments here that the main colour should be the most recent build status. How about the following layout for the "previous 5 builds", each icon would like to the respective build. 

![image](https://cloud.githubusercontent.com/assets/1097338/6050105/d32daf94-acb4-11e4-9398-00f4b3a364b8.png)

leewillis77:
Just to follow up on this - here's the layout I'm running currently. It'd need a bit of tidying up to go in, but what are everyone's thoughts on this?

![image](https://cloud.githubusercontent.com/assets/1097338/6087789/9267d1e4-ae47-11e4-80a6-5f8146f61625.png)

danaketh:
This look really nice actually.

REBELinBLUE:
That looks really good

dancryer:
I like that a lot. If you've got it working, please feel free to submit a pull request. :)

leewillis77:
PR submitted - let me know if you want any tweaks, or have any other comments!

REBELinBLUE:
How does it look on the minimal view? The one you get with more than 10 projects

leewillis77:
So, bad news: "There's a minimal view?"

Good news, I have more than 10 projects to set up ;)

I'll go check it out!

leewillis77:
The current PR doesn't touch the minimal view (Although I did spot an issue with projects with no builds not getting a highlight colour, which is resolved in the updated PR). I'd love to see the last 5 build results in the minimal view too. I'll have a play with that now.

leewillis77:
I've updated the PR to include the last 5 build details on the minimal view too:

![image](https://cloud.githubusercontent.com/assets/1097338/6098547/1411ad1c-afda-11e4-84e9-743738ac5080.png)

REBELinBLUE:
Awesome! I didn't expect many people to be aware of it, @dancryer added it when I mentioned the page was a bit big with my 30+ projects

dancryer:
I've just merged @leewillis77's new design. :+1:

PR #783, opened 2015-02-08
Dashboard layout changes.

leewillis77:
Implements design changes to the dashboard as mocked up in #708 

I'm no frontend expert, so please do review the CSS :)


dancryer:
Thanks for this @leewillis77 - The CSS all looks good, I just moved it over to AdminLTE-custom.css, but otherwise it remains unchanged.

PR #801, opened 2015-02-18
PHPCS/PHPMD fixes on Technical Debt plugin.

jfi:


PHPCI/Plugin/TechnicalDebt.php
```diff
@@ -122,21 +122,42 @@ protected function setOptions($options)
...
-    * Runs in a specified directory, to a specified standard.
+    * Runs the plugin
     */
     public function execute()
     {
         $this->phpci->logExecOutput(false);
 
-        $ignores = $this->ignore;
-        $ignores[] = 'phpci.yml';
+        list($errorCount, $data) = $this->getErrorList();
```
mavimo: You suppose that ```getErrorList``` return some values, but there you also remove return value from this function, are you sure it work? (I need to test it, but this seems an error reading code).

jfi:
Thanks @mavimo, that was a rather stupid thing to miss. :)

PR #802, opened 2015-02-18
Adding output support to Behat plugin.

jfi:
**Contribution Type:** New Feature
**Primary Area:** Plugins
**Description of change:** Adds an output block to the Behat plugin, which lists the name of each feature file and the line number of the scenario.
**Screenshot:** Screenshot

### DIFF
--- a/PHPCI/Languages/lang.en.php
+++ b/PHPCI/Languages/lang.en.php
@@ -184,6 +184,7 @@ PHPCI',
     'phpspec' => 'PHP Spec',
     'phpunit' => 'PHP Unit',
     'technical_debt' => 'Technical Debt',
+    'behat' => 'Behat',
 
     'file' => 'File',
     'line' => 'Line',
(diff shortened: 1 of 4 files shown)


jfi:
If I try run `behat -format junit --out xml` (where an `xml` directory exists), I get the following:
 
  [Behat\Testwork\Output\Exception\FormatterNotFoundException]         
  `junit` formatter is not found or has not been properly registered.  

So it appears at the very least this isn't available out-of-the-box on OS X? https://github.com/Behat/Behat/issues/358 seems to suggest it's not implemented in 3.0 whilst https://github.com/Behat/Behat/pull/676 points to it happening any day now. I'd be happy to revisit and parse the junit output (for starters, that would mean being able to display the name of the failing scenario), but for now, I think this is the best we can do - and I'd argue, better than no output at all.

jfi:
Additionally, I'm seeing an issue where (with 100ish scenarios), the Build Log is getting truncated half way through the tests - I assume due to the length of the database field. @mavimo, how would you or anyone else using Behat feel about changing the log output to be progress, rather than pretty, so this doesn't occur?

mavimo:
@jfi about log length I suppose is already fixed in #777 what version of PHPCI are you using?

jfi:
The upgrade sorted that - missed it had been merged in.

dancryer:
Mergeded.

PR #803, opened 2015-02-19
Make SSH key generation more robust.

leewillis77:
Do not try and predict whether we will be able to create a key. Instead try and create one and capture failure if it happens. Hopefully fixes #800.

We now call ssh-keygen without "-q", and capture the output. If ssh-keygen fails then shell_exec should return NULL, so the presence of output implies success.

It'd be good to get testing on multiple setups for this. 

For reference I can confirm it works on MacOS Yosemite.

PR #804, opened 2015-02-21
Add support for Github pull request comments

dancryer:
Not to be merged yet, initial support for PHPCS is done... Need to add support for: a) Other tools, b) Other platforms (Bitbucket, etc).

### DIFF
--- a/PHPCI/Helper/Github.php
+++ b/PHPCI/Helper/Github.php
@@ -108,4 +108,74 @@ class Github
 
         return $rtn;
     }
+
+    /**
+     * Create a comment on a specific file (and commit) in a Github Pull Request.
+     * @param $repo
+     * @param $pullId
+     * @param $commitId
+     * @param $file
+     * @param $line
+     * @param $comment
+     * @return null
+     */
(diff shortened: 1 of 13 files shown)


dancryer:
FYI @mavimo for comparison with your Gitlab work.

dancryer:
It seems it also needs work on detecting which diff line it should actually comment on. At the moment it assumes the first line that matches is the one, but for lines like closing braces `}`, etc. that falls down.

dancryer:
See https://github.com/Block8/PHPCI/pull/803/files?diff=unified for an example.

PHPCI/Helper/Github.php
```diff
@@ -108,4 +108,30 @@ public function getRepositories()
 
         return $rtn;
     }
+
+    public function createPullRequestComment($repo, $pullId, $commitId, $file, $line, $comment)
+    {
+        $token = Config::getInstance()->get('phpci.github.token');
```
mavimo: This need to expose the github token in project, I think we can consider to check an ENV variable and if not set check for the config, maybe this improvement can be in b8framework (let it make *twelve factor app* compliant)

PR #815, opened 2015-02-24
Fix username style in user panel block.

zviryatko:
My user name is to long and jumps to the next line.
Before:
before
After:
after


tvbeek:
:+1:

jfi:
:+1: from me too. I don't have time to merge this in today, but go ahead @tvbeek.

Issue #816, opened 2015-02-24
Archived button in list stay archived if you are in the archived list


tvbeek:
If you select Archived you get the list with the Archived projects but still the Archived button and not a button to go back to the normal list

PR #825, opened 2015-02-25
Fix archive link.

zviryatko:
Contribution Type: bug fix
Primary Area: front-end
Link to Bug: #816

Description of change: added check if url containt `archived` parameter

Description of solution: set archive link to frontpage if we already on archive page and add class active to link.

### DIFF
--- a/PHPCI/View/layout.phtml
+++ b/PHPCI/View/layout.phtml
@@ -137,7 +137,7 @@
 
             <!-- sidebar menu: : style can be found in sidebar.less -->
             <ul class="sidebar-menu">
-                <li class="active">
+                <li<?php print (array_key_exists('archived', $_GET) ? '' : ' class="active"'); ?>>
                     <a href="<?php print PHPCI_URL; ?>">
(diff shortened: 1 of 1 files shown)


PHPCI/View/layout.phtml
```diff
@@ -242,8 +242,8 @@
                     </li>
                 <?php endif; ?>
 
-                <li class="archive">
-                    <a href="<?php print PHPCI_URL; ?>?archived">
+                <li class="archive<?php print (array_key_exists('archived', $_GET) ? ' active' : '') ?>">
+                    <a href="<?php print PHPCI_URL . (array_key_exists('archived', $_GET) ? '' : '?archived'); ?>">
```
mavimo: This generate !toggling feature! on archive limk and this can be not understandable from user point of view.

mavimo:
If we add the active class on this menu item, we need also to remove from the dashboard link, otherwise we have two different active menu items.

tvbeek:
Merged

Issue #750, opened 2015-01-15
Github pull request in private repos


gumeniukcom:
Private repos have some issues:

1) Need set special github token for getting repos ( each project need own token )
2) When creating build from pull request,  we should set branch as "pr/<pull request id>", but not "master".

Issue #559, opened 2014-08-14
github pull request build


Stricted:
hello
i try to test build pull requests on a private repo and have the problem that the clone_url is https://github.com/user/repo
on private repos that is not working here we need the ssh_url to clone and that is not supported

can you please add support for private repositorys?

MartinSadovy:
Guys, please could you test https://github.com/Block8/PHPCI/pull/832

mikebronner:
This should be fixed now. Closing. (Please re-open if you still are experiencing this issue.)

PR #832, opened 2015-02-26
Github: Support pull private repository from pull request

MartinSadovy:
Solves #559, #750

### DIFF
--- a/PHPCI/Model/Build/GithubBuild.php
+++ b/PHPCI/Model/Build/GithubBuild.php
@@ -125,7 +125,7 @@ class GithubBuild extends RemoteGitBuild
 
         if ($this->getExtra('build_type') == 'pull_request') {
             $matches = array();
-            preg_match('/\/([a-zA-Z0-9_\-]+\/[a-zA-Z0-9_\-]+)/', $this->getExtra('remote_url'), $matches);
+            preg_match('/[\/:]([a-zA-Z0-9_\-]+\/[a-zA-Z0-9_\-]+)/', $this->getExtra('remote_url'), $matches);
 
             $reference = $matches[1];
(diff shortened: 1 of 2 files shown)


PHPCI/Model/Build/GithubBuild.php
```diff
@@ -125,7 +125,7 @@ public function getFileLinkTemplate()
 
         if ($this->getExtra('build_type') == 'pull_request') {
             $matches = array();
-            preg_match('/\/([a-zA-Z0-9_\-]+\/[a-zA-Z0-9_\-]+)/', $this->getExtra('remote_url'), $matches);
+            preg_match('/[\/:]([a-zA-Z0-9_\-]+\/[a-zA-Z0-9_\-]+)/', $this->getExtra('remote_url'), $matches);
```
dancryer: What does this particular change do?

dancryer:
Looks good. Would be great if someone else could test as I don't have any private github repos...

PHPCI/Model/Build/GithubBuild.php
MartinSadovy: webhook:
```json
        "ssh_url": "git@github.com:Block8/PHPCI.git",
        "clone_url": "https://github.com/Block8/PHPCI.git",
```
it is char before `Block8/PHPCI`

ViGo5190:
it's work!

LewisW:
I don't think this completely fixes #750, in my experience the problem is caused by the usage of deploy keys. It appears a deploy key is only valid for the main repository, so when people submit a pull request and it tries to fetch from the forked repository, it cannot. 

The fix is to supply your own private key which has permission the main repository and any forks, but it's a painful road of debugging and digging around to work out this is the fix.

Issue #436, opened 2014-05-18
Wrong buildPath variable


devzorg:
PHPCI/Builder.php#L293

```
        $this->buildPath = $this->ciDir . 'build/' . $buildId . '/';
```

`$this->buildPath` returns just '/'
But `$this->ciDir` returns full path (exists).
My builds can't be built cause this error.

PR #834, opened 2015-02-27
Added a variable to configure the build root.

Adirelle:
This variable allows to change where the builds are done, e.g. builds while be created in PHPCI_BUILD_ROOT_DIR/%BUILD%/. To avoid breaking existing installation, it defaults to PHPCI_DIR.'PHPCI/build/'.

(It bothered me that the builds were done inside PHPCI's sources.)

### DIFF
--- a/PHPCI/Builder.php
+++ b/PHPCI/Builder.php
@@ -215,15 +215,6 @@ class Builder implements LoggerAwareInterface
                 $this->pluginExecutor->executePlugins($this->config, 'failure');
                 $this->buildLogger->logFailure(Lang::get('build_failed'));
             }
-
-            // Clean up:
-            $this->buildLogger->log(Lang::get('removing_build'));
-
-            $cmd = 'rm -Rf "%s"';
-            if (IS_WIN) {
-                $cmd = 'rmdir /S /Q "%s"';
-            }
-            $this->executeCommand($cmd, rtrim($this->buildPath, '/'));
         } catch (\Exception $ex) {
             $this->build->setStatus(Build::STATUS_FAILED);
             $this->buildLogger->logFailure(Lang::get('exception') . $ex->getMessage());
@@ -233,6 +224,11 @@ class Builder implements LoggerAwareInterface
         // Update the build in the database, ping any external services, etc.
         $this->build->sendStatusPostback();
         $this->build->setFinished(new \DateTime());
+
+        // Clean up:
+        $this->buildLogger->log(Lang::get('removing_build'));
+        $this->build->removeBuildDirectory();
+
         $this->store->save($this->build);
     }
 
@@ -287,7 +283,7 @@ class Builder implements LoggerAwareInterface
      */
     protected function setupBuild()
     {
-        $this->buildPath = PHPCI_BUILD_ROOT_DIR . $this->build->getId() . '/';
+        $this->buildPath = $this->build->getBuildPath() . '/';
         $this->build->currentBuildPath = $this->buildPath;
 
         $this->interpolator->setupInterpolationVars(
--- a/PHPCI/Builder.php
+++ b/PHPCI/Builder.php
@@ -287,7 +287,7 @@ class Builder implements LoggerAwareInterface
      */
     protected function setupBuild()
     {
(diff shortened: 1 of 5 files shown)


mavimo:
I remember that there are some old issue where this change generate some issue, can you please spend some mins to find it and check if all work fine?

Adirelle:
I found #509 but it seems unrelated. I also found #436 but it maybe be some confusion in consequence of #509... If you could recall what was the issues, it could help.

tobiastom:
That change would be awesome — if there are no problems. :)

Adirelle:
@mavimo , @tobiastom: I think this PR is done. I would like to hear about it.

PHPCI/Model/Build.php
```diff
@@ -217,4 +217,28 @@ public function reportError(Builder $builder, $file, $line, $message)
...
+     */
+    public function removeBuildDirectory()
+    {
+        $buildPath = $this->getBuildPath();
+
+        if (!is_dir($buildPath)) {
+            return;
+        }
+
+        exec(sprintf(IS_WIN ? 'rmdir /S /Q "%s"' : 'rm -Rf "%s"', $buildPath));
```
mavimo: You switched from ```shell_exec``` to ```exec```, is this change tested also on Windows platform?

vars.php
```diff
@@ -16,6 +16,11 @@
     define('PHPCI_BIN_DIR', PHPCI_DIR . 'vendor/bin/');
 }
 
+// Define PHPCI_BUILD_ROOT_DIR
+if (!defined('PHPCI_BUILD_ROOT_DIR')) {
+    define('PHPCI_BUILD_ROOT_DIR', PHPCI_DIR . 'PHPCI/build/');
```
mavimo: I suggest to use ```DIRECTORY_SEPARATOR``` eg:

    define('PHPCI_BUILD_ROOT_DIR', PHPCI_DIR . 'PHPCI' . DIRECTORY_SEPARATOR . 'build' . DIRECTORY_SEPARATOR);

PHPCI/Builder.php
```diff
@@ -287,7 +283,7 @@ public function interpolate($input)
      */
     protected function setupBuild()
     {
-        $this->buildPath = PHPCI_DIR . 'PHPCI/build/' . $this->build->getId() . '/';
+        $this->buildPath = $this->build->getBuildPath() . '/';
```
mavimo: I prefer to use ```DIRECTORY_SEPARATOR``` instead of ```'/'````

mavimo:
@Adirelle I'm not sure ```getBuildPath()``` and ```removeBuildDirectory()``` location should be the Build model. They are not feature of Build model, but should be in a builder or some different service.

PHPCI/Model/Build.php
Adirelle: Haven't tested it. However, both shell_exec and exec pass the command to cmd.exe on Windows, the only difference being the output handling.

Adirelle:
I think this is consistent with the Build subclasses, which handle the creation of the build directory.

PHPCI/Builder.php
MarkMaldaba: All platforms support / as a directory separator, and it is a lot easier to read if you use do it this way.

The only times you need to use DIRECTORY_SEPARATOR are (a) for display purposes, so you output in the expected format for the local OS; or (b) if you're doing a comparison between paths, where one of the paths is in local format (though in this case you're safest using realpath() on both of your inputs first, to resolve any symlinks, "/../" elements, etc. and that will tidy any separators for you anyway).

Therefore it is much better to use a literal / instead of the long winded DIRECTORY_SEPARATOR unless the path is going to be shown to the user, in which case tidying at point-of-output might be the most sensible approach anyway.

vars.php
MarkMaldaba: I disagree - see my comment on line 286.

ronanchilvers:
+1 Any progress on this?

mavimo:
Since this issue seems to be important for large amount of people i prefer to add move in read-to-merge status, and have failback issue for minor changes.

PR #835, opened 2015-02-27
Added default value in profile language select (current language)

corpsee:
Added default value in profile language select (current language)


tvbeek:

Issue #302, opened 2014-03-18
Git Clone with references / cache


adamlc:
It looks the repository is cloned from a fresh each time a test is run. This can be speeded up dramatically by caching the repository locally.

Try reading this website for more info:

http://randyfay.com/content/reference-cache-repositories-speed-clones-git-clone-reference

thinkspill:
+1 :D

Spiritvn:
Sorry for reopen this, how we do config for this to work?

thinkspill:
```
build_settings:
    clone_depth: 1 # depth of 1 is a shallow clone, remove this line to clone entire repo
```

See top of the page here: https://www.phptesting.org/wiki/Adding-PHPCI-Support-to-Your-Projects

Spiritvn:
Thanks @thinkspill, i dont really understand the clone_depth setting because when i tried, it still clone the entire repo on every build?

I really need it only build the changes, not clone...

REBELinBLUE:
`clone_depth` uses the `--depth` parameter in `git clone` it makes a shallow clone, so the cloned copy won't have any history just `HEAD`. There is no way to pull in just the files which have changed, and I'm not sure that would be desirable anyway as then things like unit tests would almost certainly fail?

Spiritvn:
@REBELinBLUE thanks, ad yes that's why i asked for the config to make the commit https://github.com/Block8/PHPCI/pull/838 to work. It's ok to build the new one to test but shouldn't clone entire source, for big project it would be so long on that part.

I just didn't notice that it was closed instead of merged as you did... ^^

PR #838, opened 2015-02-27
Use persistent repository mirrors to speed up git cloning.

Adirelle:
This is basically an implementation for https://github.com/Block8/PHPCI/issues/302.

Once the administrator sets up a directory and indicates it in config.yml, PHPCI creates mirrors of the remote repositories and reuses them in subsequence builds.

E.g. given, the configuration:

```yml
phpci:
  git:
    mirrors: /var/cache/phpci
```

When cloning https://github.com/Block8/PHPCI.git, for the first time, PHPCI will do this:

```bash
git clone --mirror "https://github.com/Block8/PHPCI.git" "/var/cache/phpci/https___github_com_Block8_PHPCI_git"
git clone --reference="/var/cache/phpci/https___github_com_Block8_PHPCI_git" --branch="master" "https://github.com/Block8/PHPCI.git" "/phpci/PHPCI/build/85"
```

On subsequence build,s it will do this:
```bash
git --git-dir="/var/cache/phpci/https___github_com_Block8_PHPCI_git" remote update --prune 
git clone --reference="/var/cache/phpci/https___github_com_Block8_PHPCI_git" --branch="master" "https://github.com/Block8/PHPCI.git" "/phpci/PHPCI/build/85"
```

### DIFF
--- a/PHPCI/Service/ProjectService.php
+++ b/PHPCI/Service/ProjectService.php
@@ -99,7 +99,7 @@ class ProjectService
         $rtn = $this->projectStore->save($project);
 
         // Cleanup if archived or the reference changed
-        if(isset($modified['archived']) || isset($modified['reference'])) {
+        if (isset($modified['archived']) || isset($modified['reference'])) {
             $project->cleanup();
         }
 
--- a/PHPCI/Service/ProjectService.php
+++ b/PHPCI/Service/ProjectService.php
@@ -92,8 +92,18 @@ class ProjectService
         // Allow certain project types to set access information:
         $this->processAccessInformation($project);
 
+        // Get modified attributes
+        $modified = $project->getModified();
+
         // Save and return the project:
-        return $this->projectStore->save($project);
+        $rtn = $this->projectStore->save($project);
+
+        // Cleanup if archived or the reference changed
+        if(isset($modified['archived']) || isset($modified['reference'])) {
+            $project->cleanup();
+        }
+
+        return $rtn;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Adirelle:
Note: add a cleanup mechanism (when the project is archived/deleted ?).

tvbeek:
I think it is important to add the cleanup mechanism before merging this one to prevent the wast of disk space on mirrors that aren't used

Adirelle:
Note sure how the cleanup should work, since the mirror paths are URL-based ; several projects using the same remote URL will reuse the same mirror, by the way. So there is two possible ways:
* Use a path based on the project identifier and clean it when the project is removed, archived or when the remote URL is changed.
* Cheaper: a cron job removing old directories.

Any thought ?

tvbeek:
maybe is it an idea to create a cache dir for each project
So not /var/cache/phpci/ but /var/cache/phpci/[project_id]

It gives a change that you cache the same repository twice but you can clean up after a removal

PHPCI/Service/ProjectService.php
```diff
@@ -92,8 +92,18 @@ public function updateProject(Project $project, $title, $type, $reference, $opti
...
         $this->processAccessInformation($project);
 
+        // Get modified attributes
+        $modified = $project->getModified();
+
         // Save and return the project:
-        return $this->projectStore->save($project);
+        $rtn = $this->projectStore->save($project);
+
+        // Cleanup if archived or the reference changed
```
phpcibot: PHPCS: Expected 1 space after IF keyword; 0 found

phpcibot: PHPCS: Expected 1 space after IF keyword; 0 found

Adirelle:
So the cleanup code is in place. However, it requires proper permissions on the mirror directory: both the phpci user (who runs the daemon/cron-job) and the web user (www-data on Debian) must have write permissions to that directory, the former to create them, the latter to remove them.

REBELinBLUE:
@dancryer Any reason this was just closed rather than merged?

PR #844, opened 2015-03-02
Fixed an inconsistency in the way the prompts in the install scripts were being output.

MarkMaldaba:
All the DB/host fields have a space after the colon, which is much better as it means the text you type is slightly separated from the prompt.  However, the admin user fields didn't include this space which was inconsistent and made the install script a little less professional.

I have therefore added the missing space for the prompts which didn't have it in the English language file, and have also updated all other language files to also use this format.  Most of them followed the same inconsistency as the English version, though some were consistent but without a space, and some which were differently inconsistent (both internally, and between languages).


tvbeek:
:+1:

PR #845, opened 2015-03-02
Fixed the install script, which bails-out with an error if the PHPCI path contains spaces.

### LINKED ISSUE #698: Impossible to install on Windows
Hey, so I've jsut decided to install PHPCI on my Windows machine and I have found this little problem, that prevents PHPCI from installing. Every time I run the `php ./console phpci:install` command, I get all the way to the `Setting up your database...` and then, this error appears: `'C:\Program' is not recognized as an internal or external command, operable program or batch file` and the installation continues, but then this appears: `SQLSTATE[42S02]: Base table or view not found: 1146 Table 'phpci.user' doesn't exist` and the installation breaks...

So what could be the problem?

MarkMaldaba:
This occurs commonly on Windows, but from my reading of the code it would also be a problem on other platforms if spaces were present (though this is less likely, due to different naming conventions).

It has been fixed by using escapeshellarg() on both of the paths used in the command.

Fixes #698, which I've just noticed has a similar solution suggested in one of the comments, but was closed without anyone actually implementing it.

### DIFF
--- a/PHPCI/Command/InstallCommand.php
+++ b/PHPCI/Command/InstallCommand.php
@@ -323,8 +323,8 @@ class InstallCommand extends Command
     {
         $output->write(Lang::get('setting_up_db'));
 
-		$phinxBinary = escapeshellarg(PHPCI_DIR . 'vendor/bin/phinx');
-		$phinxScript = escapeshellarg(PHPCI_DIR . 'phinx.php');
+        $phinxBinary = escapeshellarg(PHPCI_DIR . 'vendor/bin/phinx');
+        $phinxScript = escapeshellarg(PHPCI_DIR . 'phinx.php');
         shell_exec($phinxBinary . ' migrate -c ' . $phinxScript);
 
         $output->writeln('<info>'.Lang::get('ok').'</info>');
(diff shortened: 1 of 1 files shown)


tvbeek:
:+1:

PHPCI/Command/InstallCommand.php
```diff
@@ -323,7 +323,9 @@ protected function setupDatabase(OutputInterface $output)
     {
         $output->write(Lang::get('setting_up_db'));
 
-        shell_exec(PHPCI_DIR . 'vendor/bin/phinx migrate -c "' . PHPCI_DIR . 'phinx.php"');
+		$phinxBinary = escapeshellarg(PHPCI_DIR . 'vendor/bin/phinx');
```
mavimo: Use space for indentation.

MarkMaldaba:
Has this not been merged because it has the 'requires-changes' flag?  I fixed the issues almost immediately, but it looks like only admins can turn the flag off, and if admins are ignoring the pull request that require changes then we are in a bit of a catch-22 situation!

It is a simple fix for a critical problem (which stops PHPCI from working on Windows) so I don't see what the hold up is, otherwise.

mavimo:
@MarkMaldaba people can have time but not enough to re-review all the issue/PR every time, sometime after the first review the review list of the issue "push down" the current MR and "disappear" from the MR list. Sometime you can add a comment with users involved to require a new CR after the fixes are done.

MarkMaldaba:
@mavimo it would be good if there were a way to indicate that the required changes have been implemented, but thanks for the merge.

PR #849, opened 2015-03-04
Flexible layout scheme in Subversion

minderm:
As Subversion doesn't constrain repository layout and users can run any layout on their systems, this pull request adds the flexibility providing users the ability to specify concrete folders paths in the project configuration in case when their layout differs from "project/trunk, project/branches, project/tags" scheme. 

The project configuration example for this case:

svn:
    username:  some_user
    password: xxx
    trunk-path: trunk/big/fat/project
    branches-path: branches/big/fat/project


tvbeek:
It looks good but I don't have the option to check it
Does any @Block8/phpci-contributors have that option?

dancryer:
@minderm This looks great, thanks for fixing up my hacked-in changes. :+1: 

Please can you rebase this on top of the current master branch? It doesn't want to apply cleanly for me at present.

Other than that, good to merge from me.

PR #854, opened 2015-03-05
Symlinks removed incorrectly. Fixed.

vsguts:
For a project of type "Local Path" with the "prefer_symlink: true," when the build is removed (i.e. completed), not the symlink was removed, but the location it points to, i.e. the original repository.


tvbeek:
:+1:

PR #856, opened 2015-03-06
Minor improvements in GitHub webhook and CSS

rene-s:
Hi, I have tried to trigger a build by calling PHPCI's GitHub webhook URL and assumed it would work. At first I was confident since PHPCI returned "...OK...". It did not work though, there was a PHP error message in php_error.log stating that $payload was null when it should have been an array.

The second commit actually loads the PHPCI logo when the app has been installed in a sub directory.

Bye
Rene


tvbeek:
If I'm right you have one pull request with multiple changes. Can you split it in multiple pull requests?

PR #858, opened 2015-03-07
archived

Lechus:


tvbeek:
Can you add a description?

@dancryer Can you the phpci installation because there isn't any response

dancryer:
@tvbeek It is restarting now... but as this is a one-line change to add a translation, I don't think we need to wait for a response. :+1:

tvbeek:

PR #861, opened 2015-03-09
Added dbunit to PHPCI's composer

vsguts:
PHPCI comes with its own version of phpunit, but doesn't include dbunit.

My local phpunit installation includes dbunit by default. Because of this, tests would run locally but fail remotely when executed by PHPCI.

I added dbunit to composer, so that PHPCI's phpunit would run tests like a regular phpunit installation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
 
     "require-dev": {
         "phpunit/phpunit": "~4.0",
+        "phpunit/dbunit": ">=1.2",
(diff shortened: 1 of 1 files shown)


dancryer:
Hi @vsguts 

You need to add dbunit to your own project's `composer.json` file, in the `require-dev` section.

PHPCI looks in your project's composer directory before it looks at its own. Dbunit isn't a requirement for PHPCI, so we couldn't include it in our standard config.

PR #864, opened 2015-03-11
Refactored the run-builds, rebuild and daemonise commands.

Adirelle:
This a refactorisation of these commands, it consists of the following modifications:
* The former RunCommand->execute method has been split into smaller methods: setupLogging, findNextBuild and runBuild.
* I got rid of all instance attributes but $logger; they were not necessary and created some confusion.
* RunCommand->runBuild *always* pops the build logger and *always* removes the build directory.
* DaemoniseCommand and RebuildCommand now inherits from RunCommand instead of creating an instance; their execute method simply reuses the new methods to implement a different logic.

The command logic is now:
* RunCommand: 
  * Runs up to maxBuilds (default 100) builds, one at a time. 
  * Verifies running builds and looks for a pending build at each iteration (instead of finding all pending builds at the start).
  * Ends as soon as there is nothing to do. This fixes an issue where the RunCommand goes in a very short loop, doing requests and spitting in the log files at a very fast pace.
* DaemoniseCommand:
  * Runs indefinitely.
  * Sleeps where there is nothing to do, using an incremental duration.
* RebuildCommand:
  * Duplicates the latest build and build it immediately, bypassing all restrictions.

I would like to add POSIX process control (handling signals, tracking build PIDs and killing them when they time out) but that would not work on Windows; another solution could be to use the proc_* family of functions to do so. Any thought on this ?


PHPCI/Command/DaemoniseCommand.php
```diff
@@ -65,46 +40,30 @@ protected function configure()
...
 
-            try {
-                $buildCount = $runner->run($emptyInput, $output);
-            } catch (\Exception $e) {
-                $output->writeln('<error>Exception: ' . $e->getMessage() . '</error>');
-                $output->writeln('<error>Line: ' . $e->getLine() . ' - File: ' . $e->getFile() . '</error>');
+            if ($build) {
+                $this->runBuild($build);
+                $sleep = 0;
+                echo 'build';
```
dancryer: Just a quick one - Anywhere that you're currently using `echo`, can you please switch it over to using `$output->writeln(...)`

dancryer:
This looks great. Big improvement. :+1:

PHPCI/Command/DaemoniseCommand.php
Adirelle: I have kept these "echo"s from the previous version but I don't really like them anyway. I'll change them.

PHPCI/Command/RunCommand.php
```diff
@@ -164,7 +180,7 @@ protected function validateRunningBuilds()
             if (($now - $start) > $timeout) {
                 $this->logger->addInfo(Lang::get('marked_as_failed', $build->getId()));
                 $build->setStatus(Build::STATUS_FAILED);
-                $build->setFinished(new \DateTime());
+                $build->setFinished(new DateTime());
```
mavimo: Since we arenìt in the root namespace we need the slash before DateTime object.

PHPCI/Command/RunCommand.php
```diff
@@ -65,89 +54,116 @@ protected function configure()
...
+
+        $this->removeBuildDirectory($build);
+        $this->logger->popHandler($buildDbLog);
+        $this->logger->addInfo("Build ".$build->getId()." ended");
     }
 
+    /**
+     * Checks all running builds, and kills those that seem dead.
+     *
+     * @return array An array of projects having running builds.
```
mavimo: Since we return a project array return type definition should be ```Project[]```

PHPCI/Command/RunCommand.php
```diff
@@ -65,89 +54,116 @@ protected function configure()
...
+        } catch (Exception $ex) {
+            $build->setStatus(Build::STATUS_FAILED);
+            $build->setFinished(new DateTime());
+            $build->setLog($build->getLog() . PHP_EOL . PHP_EOL . $ex->getMessage());
+            Factory::getStore('Build')->save($build);
+        }
+
+        $this->removeBuildDirectory($build);
+        $this->logger->popHandler($buildDbLog);
+        $this->logger->addInfo("Build ".$build->getId()." ended");
```
mavimo: Please use sprintf for string formatting, this will help the translation step.

```php
        $this->logger->addInfo(sprintf("Build %d ended", $build->getId()));
```

PHPCI/Command/RunCommand.php
```diff
@@ -65,89 +54,116 @@ protected function configure()
...
+     * Runs one build and cleans up after it finishes.
+     *
+     * @param Build $build
+     */
+    protected function runBuild(Build $build)
     {
-        $this->maxBuilds = (int)$numBuilds;
-    }
+        $build = BuildFactory::getBuild($build);
+        $this->logger->addInfo("Running build ".$build->getId());
```
mavimo: Use sprintf for string formatting (see below)

PHPCI/Command/RunCommand.php
```diff
@@ -65,89 +54,116 @@ protected function configure()
...
+            } else {
+                return $build;
             }
-
         }
 
-        $this->logger->addInfo(Lang::get('finished_processing_builds'));
-
-        return $builds;
+        return null;
```
mavimo: ```return null``` is not required.

PHPCI/Command/DaemoniseCommand.php
```diff
@@ -65,46 +40,27 @@ protected function configure()
...
-        exec($command);
+        $this->setupLogging($input, $output);
 
-        $this->output = $output;
-        $this->run   = true;
-        $this->sleep = 0;
-        $runner      = new RunCommand($this->logger);
-        $runner->setMaxBuilds(1);
-        $runner->setDaemon(true);
+        file_put_contents(PHPCI_DIR."/daemon/daemon.pid", getmypid());
```
mavimo: User const for directory separator instead of /

mavimo:
Really big improvement, AWESOME!

Before merge I think we need to add tests to cover this feature.

Adirelle:
Adding unit tests is quite tricky because there are a lot of hard-coded dependencies in these classes (as well as all over PHPCI, btw). Those hard-coded dependencies prevent using mocks.

mavimo:
@Adirelle I know, but we need to start from some point, and this is an important feature of PHPCI, so I think need to be tested. Some sample for command should be visibile in [Tests/PHPCI/Command](https://github.com/Block8/PHPCI/blob/master/Tests/PHPCI/Command/CreateAdminCommandTest.php).

Adirelle:
The issue is the calls to static methods, which cannot be mocked. Here is an explanation why: http://misko.hevery.com/2008/12/15/static-methods-are-death-to-testability/ Said otherwise, you cannot unit test a code that uses hard dependencies and global state unless you refactor it to be fully injectable.

Note that in this case, it is possible by extracting the running logic from the command to create a standalone class.

PHPCI/Helper/Runner.php
```diff
@@ -0,0 +1,179 @@
...
+
+            if (IS_WIN) {
+                $cmd = 'rmdir /S /Q "%s"';
+            }
+
+            exec($cmd);
+        }
+    }
+
+}
```
phpcibot: PHPCS: The closing brace for the class must go on the next line after the body

Adirelle:
* Rebased on latest master.
* Extracted most of logic in an helper class, Runner. Unit tests are planned.
* Also reworked the daemon and daemonise commands to use the posix and pcntl extensions (which makes sense since you cannot "daemonise" on non-POSIX OSes like Windows) instead of relying on the shell.

PHPCI/Command/DaemonCommand.php
```diff
@@ -40,85 +49,133 @@ protected function configure()
...
 
     /**
-    * Loops through running.
-    */
+     * Loops through running.
+     */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $state = $input->getArgument('state');
+        if(!function_exists('pcntl_fork') || !function_exists('posix_setsid')) {
```
dancryer: I'm not sure I am comfortable with adding a new dependency (pcntl) for something that worked fine before (without the dependency).

PHPCI/Helper/Runner.php
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * @var integer
+     */
+    protected $buildTimeout;
+
+    /**
+     * @param LoggerInterface $logger
+     * @param string $name
+     */
+    public function __construct(BuildFactory $builderFactory, LoggerInterface $logger, $buildStore, $buildTimeout = 1800)
```
phpcibot: PHPCS: Line exceeds 120 characters; contains 121 characters

PHPCI/BuilderFactory.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    /**
+     * @var LoggerInterface
+     */
+    protected $logger;
+
+    /**
+     * @param LoggerInterface $logger
+     */
+    function __construct(LoggerInterface $logger)
```
phpcibot: PHPCS: Visibility must be declared on method "__construct"

PHPCI/Builder.php
```diff
@@ -404,4 +396,22 @@ function () use ($self) {
 
         return $pluginFactory;
     }
+
+    /**
+     * Remove the build directory of a finished build.
+     */
+    public function removeBuildDirectory()
+    {
+        if(!is_dir($this->buildPath)) {
```
phpcibot: PHPCS: Expected 1 space after IF keyword; 0 found

Adirelle:
I finally managed to have something testable without much hassle.

The decisive points were:
* Delegate the Builder creation to a factory,
* Have the constructors accept optional dependencies, building the missing ones. This is used to inject mocks in the tests.

I still wonder where removeBuildDirectory should go. Builder is nice candidate, but it cannot be used to prune timed-out builds.

And there is still no way to check if a build is really running or stuck, nor to interrupt a build that runs for too long.

mavimo:
RUNNING PLUGIN: php_mess_detector

Exception: Notice: Undefined offset: 28 in /phpci/PHPCI/Model/Build/GithubBuild.php line 224

Adirelle:
It seems to be a bug in the PHPCI instance building the test... since I haven't touched GithubBuild in that PR.

Adirelle:
Found it :

```
PHPCI/Builder.php	28	417	The class Builder has 11 methods. Consider refactoring Builder to keep number of methods under 10. 
```

Adirelle:
Considering several issues I've seen recently, I wondering about slightly rewriting this PR this way:
* Have the phpci:daemonize commands just runs "./daemonise phpci:run-builds; sleep 15" in a infinite loop, without ever looking at the database. This should fix the issues about Project caching and stale Mysql connections.
* Have the phpci:run-builds command run each build in a separate sub-command. This should ensure build isolation and we could eventually track the builds using PIDs. However, we would run into trouble if run-builds had to modify the Build object after launching the sub-command, as it would work with a stale Build object.
* Maybe add a ProcessControl helper with two methods -- isRunning($pid) and kill($pid, $signal) -- and different implementations depending on installed extensions and the OS, to help tracking and checking processes in the previous rewrites.

Should I close this PR and open a new one or rewrite it ?

Adirelle:
After working on other parts of PHPCI and given some bugs, I think the these commands should be reworked using another approach. Moreover, this would probably be integrated into other PR, e.g. #884.

PR #873, opened 2015-03-17
Feature login by name/email

corpsee:
Improved login: now you can login using name or email


mavimo:
In order to work we need to ensure that e-mail is unique, currently is not, we need to create a new migration to create a unique index on email field, see [phinix](http://phinx.readthedocs.org/en/latest/migrations.html#working-with-indexes) docs.

PHPCI/Migrations/20150324174958_fix_email_and_name_user_fields.php
```diff
@@ -0,0 +1,30 @@
...
+class FixEmailAndNameUserFields extends AbstractMigration
+{
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $user_table = $this->table('user');
+        $user_table
+            ->addIndex('email', ['unique' => true])
```
mavimo: Use ```array()``` instead of ```[]``` in order to support PHP 5.3. Same change on the next lines..

PHPCI/Migrations/20150324174958_fix_email_and_name_user_fields.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class FixEmailAndNameUserFields extends AbstractMigration
```
mavimo: I prefer a different name for class and file, ```unique``` should be nbetter then ```fix```

PHPCI/Migrations/20150324174958_fix_email_and_name_user_fields.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    public function down()
+    {
+        $user_table = $this->table('user');
+        $user_table
+            ->removeIndex('email', ['unique' => true])
+            ->removeIndex('name', ['unique' => true])
+            ->save();
+    }
+}
```
mavimo: I prefer to add an empty line to the end of the files, but I'm not sure CS require it...

corpsee:
@mavimo, fixed

PR #882, opened 2015-03-25
Add the --recursive parameter to the git clone to get the submodules

tvbeek:
With the recursive parameter you can also get the submodules for the project


tvbeek:

Issue #867, opened 2015-03-12
PHPCI ignores branches that have been "fast-forwarded"


Adirelle:
In our team, we develop mainly on the "develop" branch. When it is ready to test, we merge develop into the "test" branch, which generally results in a fast-forward, and then push them both to git. When the "test" branch build passes, it should run a deploy script to push the build on our test server.

However, the webhook automatically depuplicates builds based solely on project and commit identifiers, see https://github.com/Block8/PHPCI/blob/master/PHPCI%2FController%2FWebhookController.php#L319. So when we push the "test" branch to the server, it points at the same commit as "develop" because of the "fast-forward", and the build is ignored by PHPCI.

Could we change this behavior to ignore hooks based on the trippler (project, branch, commit) instead ?

tvbeek:
Maybe we can add a check on commit hash (and make it an option to set to skip it) and/or combine it with #656

Adirelle:
By "commit identifiers" I meant "commit hashes".

PR #883, opened 2015-03-26
Detailled webhook responses

Adirelle:
Add details to webhook responses, namely:
* Add the build id when a build is actually created in response to the hook,
* Tell if the build is ignored, and for which reason,
* For hooks submitting several commits, collect the results of each commit in a "commit" object.

### DIFF
--- a/PHPCI/Controller/WebhookController.php
+++ b/PHPCI/Controller/WebhookController.php
@@ -338,8 +338,14 @@ class WebhookController extends \b8\Controller
      *
      * @throws \Exception
      */
-    protected function createBuild(Project $project, $commitId, $branch, $committer, $commitMessage, array $extra = null)
-    {
+    protected function createBuild(
+        Project $project,
+        $commitId,
+        $branch,
+        $committer,
+        $commitMessage,
+        array $extra = null
+    ) {
         // Check if a build already exists for this commit ID:
         $builds = $this->buildStore->getByProjectAndCommit($project->getId(), $commitId);
 
--- a/PHPCI/Controller/WebhookController.php
+++ b/PHPCI/Controller/WebhookController.php
@@ -55,7 +55,6 @@ class WebhookController extends \PHPCI\Controller
     public function bitbucket($project)
     {
         $response = new b8\Http\Response\JsonResponse();
-        $response->setContent(array('status' => 'ok'));
 
         $payload = json_decode($this->getParam('payload'), true);
 
@@ -65,7 +64,9 @@ class WebhookController extends \PHPCI\Controller
                 $email = substr($email, 0, strpos($email, '>'));
                 $email = substr($email, strpos($email, '<') + 1);
 
-                $this->createBuild($project, $commit['raw_node'], $commit['branch'], $email, $commit['message']);
+                $this->response->setContent(
+                    $this->createBuild($project, $commit['raw_node'], $commit['branch'], $email, $commit['message'])
+                );
             } catch (\Exception $ex) {
                 $response->setResponseCode(500);
                 $response->setContent(array('status' => 'failed', 'error' => $ex->getMessage()));
@@ -84,7 +85,6 @@ class WebhookController extends \PHPCI\Controller
     public function git($project)
     {
         $response = new b8\Http\Response\JsonResponse();
-        $response->setContent(array('status' => 'ok'));
 
         $branch = $this->getParam('branch');
         $commit = $this->getParam('commit');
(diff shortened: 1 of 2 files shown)


mavimo:
@Adirelle I prefer that internal methods like:

    $this->githubPullRequest($project, $payload, $response);

don't return a response but an array of information and then the the result is put in a response, so we have public methods that contains response and populate it with the appropriate informations and protected methods return informations to be put in the response.

Adirelle:
This seems fair.

Adirelle:
Actually we could also override Controller::handleAction to encode the response (or exception) in Json and have the actions only returns an array.

mavimo:
@Adirelle: should be a nice idea to clean the code on actions!

```php
    /**
     * Handles an action on this controller and returns a Response object.
     * 
     * Controller actions should return an array with information need to be 
     * send as response; if the return array have the key *reseponseCode*
     * this information is used to set the HTTP response code.
     *
     * @param string $action
     * @param array $actionParams
     *
     * @return b8\Http\Response
     */
    public function handleAction($action, $actionParams)
    {
        $response = new b8\Http\Response\JsonResponse();

        try {
            // Invoke the appropriate action
            $data = parent::handleAction($action, $actionParams);

            // If result have a response code use this information in the result.
            if (isset($data['reseponseCode'])) {
                $response->setResponseCode($data['reseponseCode']);
                unset($data['reseponseCode']);
            }

            // Set content on the response
            $response->setContent($data);
        } catch (\Exception $e) {
            // Wrap all errors in actions, remove error management from actions.
            $response->setResponseCode(500);
            $response->setContent(array('status' => 'failed', 'error' => $ex->getMessage()));
        }

        return $response;
    }
```

then actions can be simplified. like:

```php
    // ...

    /**
     * Called by POSTing to /webhook/git/<project_id>?branch=<branch>&commit=<commit>
     *
     * @param string $project
     */
    public function git($project)
    {
        $branch = $this->getParam('branch', 'master');
        $commit = $this->getParam('commit', null);
        $commitMessage = $this->getParam('message', null);
        $committer = $this->getParam('committer', null);

        return $this->createBuild($project, $commit, $branch, $committer, $commitMessage);
    }

    // ...

    /**
     * Called by Gitlab Webhooks:
     */
    public function gitlab($project)
    {
        $payloadString = file_get_contents("php://input");
        $payload = json_decode($payloadString, true);

        // build on merge request events
        if (isset($payload['object_kind']) && $payload['object_kind'] == 'merge_request') {
            $attributes = $payload['object_attributes'];
            if (in_array($atributes['state'], array('opened', 'reopened')) {
                return $this->createBuild(
                    $project,
                    $attributes['last_commit']['id'],
                    $attributes['source_branch'],
                    $attributes['last_commit']['author']['email'],
                    $attributes['last_commit']['message']
                );
            }
        }
        // build on push events
        if (isset($payload['commits']) && is_array($payload['commits'])) {
            // If we have a list of commits, then add them all as builds to be tested:
            $results = array();
            foreach ($payload['commits'] as $commit) {
                $branch = str_replace('refs/heads/', '', $payload['ref']);
                $results[$commit['id']] = $this->createBuild(
                    $project,
                    $commit['id'],
                    $branch,
                    $commit['author']['email'],
                    $commit['message']
                );
            }
            return array('status' => 'ok', 'commits' => $results);
        }

        return array('status' => 'failed', 'error' => 'Invalid payload'));
    }

    // ...
```

PHPCI/Controller/WebhookController.php
```diff
@@ -249,83 +264,90 @@ protected function githubPullRequest($projectId, array $payload, b8\Http\Respons
...
+                    $results[$commit['id']] = $this->createBuild(
+                        $project,
+                        $commit['id'],
+                        $branch,
+                        $committer,
+                        $commit['message']
+                    );
+                    $status = 'ok';
+                } catch (\Exception $ex) {
+                    $results[$commit['id']] = array('status' => 'failed', 'error' => $ex->getMessage());
```
phpcibot: PHPCS: Line exceeds 120 characters; contains 121 characters

Adirelle:
WIP. This needs tests.

I'm wondering about fixing #867 by the way.

PHPCI/Controller/WebhookController.php
Adirelle: There is a bug in the diff calculator: this errors is line 341, not 293 or 309.

tvbeek:

tvbeek:

Issue #885, opened 2015-03-26
Intent to implement: build-specific environment


Adirelle:
### What's the idea ?

Create a class which holds the build environment.

It would be fed with the default environment (maybe filtered, see below), build data, and then shared with several components:
* the Builder would add the main vendor/bin to the PATH.
* the BuildInterpolator would use it for the interpolations.
* the CommandExecutor would:
  1. use it as the the command environment,
  1. use the PATH variable to look for binairies.
* the Composer plugin would add the project vendor bin to the PATH (which synergizes with the previous point).
* the Env plugin would simply adds new variables into it (which synergizes with the BuildInterpolator).
* any plugin that would need to use it.

I'm considering filtering $_ENV to keep only meaningful variables, e.g. PATH, USER, HOME, LANG (or not), TERM (or setting it to "ansi-generic"), but determining the minimum set of *absolutely required* variables depending on the OS isn't trivial.

We could also add some options in the main configuration file to define some variables.

### Why ?

Right now, the environment of the main PHP process is used for every build, with no cleanup between build. So variables added by the Env plugin bleed from one build to another. Moreover, the original environment of the phpci:run-builds or phpci:daemonise commands isn't filtered. Using a filtered would allow to have a more isolated build process.

By using a separate object for each build, we would be sure to isolate each of them. Moreover, using an object instead of an array would allow to share its state amongst different plugins and components.

### Drawback ?

For this to work, *every* command of *every* build steps and plugins should pass through the CommandExecutor, which would probably lead to small changes in some plugins that uses "exec" directly.

dancryer:
I like the sound of this. :+1: 

My only query is this bit: `use the PATH variable to look for binairies.` - That's effectively what we do already by using `which <binary>` (or on Windows, `where <binary>`). I'd expect that replacing that with something that traverses the PATH variable would be more fragile and prone to unexpected errors.

Adirelle:
> My only query is this bit: use the PATH variable to look for binairies. - That's effectively what we do already by using which <binary> (or on Windows, where <binary>). I'd expect that replacing that with something that traverses the PATH variable would be more fragile and prone to unexpected errors.

Well, `which binary`/`where binary` would use the PATH anyway. It also worths noting that, in relation with #884, symfony/process also provides a search bina

Adirelle:
WIP there : https://github.com/Adirelle/PHPCI/tree/shared-environment

Adirelle:
[...]

PR #888, opened 2015-03-28
Shared environment

Adirelle:
Implementation of #885. I reworked the PR commits several times to have them more readable.

I ran into an issue though: unless the PHP setting "variable_order" contains "E", the $_ENV is empty so it cannot be used as a base environment. Unfortunately, this seems to be the case by default. As a workaround, the Environment class import a limited list of well-known variables.

I hesitate to append PHPCI_DIR and PHPCI_BIN_DIR to the PATH. This would allow to greatly simplify BaseCommandExecutor::findBinary as both the "which" and "where" commands accepts several arguments.


Adirelle:
Apparently there is an issue with PHPCI\Plugin\Tests\Helper\EnvironmentTest::testAddBuildVariables when executed from PHPCI but I can reproduce it locally, and I don't have the error output.

PR #891, opened 2015-03-29
Provide widget visibility control.

zviryatko:
Primary Area: front-end

Description of change: Provide ability to roll up/down widgets in build view and save it to cookies.

phpci-feature-widget-visibility

### DIFF
--- a/PHPCI/View/layout.phtml
+++ b/PHPCI/View/layout.phtml
@@ -30,6 +30,7 @@
     </script>
 
     <script src="//cdnjs.cloudflare.com/ajax/libs/jquery/2.1.1/jquery.js"></script>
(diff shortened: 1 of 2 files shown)


LewisW:
The order for the plugins uses localStorage, it might be an idea to change yours to also use local storage for consistency.

Also I personally think this would be great combined with the ability to expand the widget to force full-width - as a lot of the time the widget is too small. I'd be interested to hear people's thoughts.

zviryatko:
LocalStorage it's good idea, i did it with cookie 'cause doesn't find any other functionality.
Btw, order settings is doesn't work for me(

mikebronner:
@zviryatko Agree with @LewisW , anything that is configurable to the user on the build page should be in LocalStorage. Also please rebase with changes from `mater`. Thanks! :)

PR #892, opened 2015-03-30
Error handler fixes

Adirelle:
A couple of changes to ensure the error/exception handler is registered and forwards the error logs to any logger created by LoggerConfig::getFor.

Moreover, LoggerConfig::getFor always returns the same instance of Logger for the same $name. This ensures calls to pushHandler & pushProcessor are not "lost" between successive calls to getFor.

### DIFF
--- a/PHPCI/Logging/LoggerConfig.php
+++ b/PHPCI/Logging/LoggerConfig.php
@@ -19,6 +19,7 @@ class LoggerConfig
 {
     const KEY_ALWAYS_LOADED = "_";
     private $config;
+    private $cache = array();
 
     /**
      * The filepath is expected to return an array which will be
@@ -56,9 +57,20 @@ class LoggerConfig
      */
     public function getFor($name)
     {
+        if (isset($this->cache[$name])) {
+            return $this->cache[$name];
+        }
+
         $handlers = $this->getHandlers(self::KEY_ALWAYS_LOADED);
-        $handlers = array_merge($handlers, $this->getHandlers($name));
-        return new Logger($name, $handlers);
+        if ($name !== self::KEY_ALWAYS_LOADED) {
+            $handlers = array_merge($handlers, $this->getHandlers($name));
+        }
+
+        $logger = new Logger($name, $handlers);
+        Handler::getInstance()->register($logger);
+        $this->cache[$name] = $logger;
+
+        return $logger;
     }
 
     /**
--- a/PHPCI/Logging/LoggerConfig.php
+++ b/PHPCI/Logging/LoggerConfig.php
@@ -9,6 +9,7 @@
 
 namespace PHPCI\Logging;
 
+use Monolog\ErrorHandler;
 use Monolog\Logger;
 
 /**
@@ -67,7 +68,7 @@ class LoggerConfig
         }
 
         $logger = new Logger($name, $handlers);
-        Handler::getInstance()->register($logger);
+        ErrorHandler::register($logger);
         $this->cache[$name] = $logger;
 
         return $logger;
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -8,7 +8,6 @@
 */
 
 // Let PHP take a guess as to the default timezone, if the user hasn't set one:
-use PHPCI\Logging\Handler;
 use PHPCI\Logging\LoggerConfig;
 
 $timezone = ini_get('date.timezone');
@@ -43,9 +42,10 @@ if (!file_exists(dirname(__FILE__) . '/vendor/autoload.php') && defined('PHPCI_I
 // Load Composer autoloader:
 require_once(dirname(__FILE__) . '/vendor/autoload.php');
 
+\PHPCI\ErrorHandler::register();
+
 if (defined('PHPCI_IS_CONSOLE') && PHPCI_IS_CONSOLE) {
     $loggerConfig = LoggerConfig::newFromFile(__DIR__ . "/loggerconfig.php");
-    Handler::register($loggerConfig->getFor('_'));
 }
 
 // Load configuration if present:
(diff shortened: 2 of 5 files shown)


PHPCI/Logging/Handler.php
```diff
@@ -33,30 +53,39 @@ class Handler
...
-     * @param LoggerInterface $logger
+     * @var boolean
      */
-    public function __construct(LoggerInterface $logger = null)
+    protected $registered = false;
+
+    /**
+     * @param Logger $logger
+     */
+    public function __construct(Logger $logger = null)
```
mavimo: Should be ```LoggerInterface```?

Adirelle: Nope: \Monolog\Logger, because pushHandler (line 87) is a method of \Monolog\Logger.

Adirelle:
@mavimo: I think this PR is done. Could you review it please ?

PHPCI/Logging/LoggerConfig.php
```diff
@@ -56,9 +57,19 @@ public function __construct(array $configArray = array())
      */
     public function getFor($name)
     {
-        $handlers = $this->getHandlers(self::KEY_ALWAYS_LOADED);
-        $handlers = array_merge($handlers, $this->getHandlers($name));
-        return new Logger($name, $handlers);
+        if (!isset($this->cache[$name])) {
```
mavimo: I prefer the return first approach:

```php
    public function getFor($name)
    {
        if (isset($this->cache[$name])) {
            return $this->cache[$name];
        }

        $handlers = $this->getHandlers(self::KEY_ALWAYS_LOADED);
        if ($name !== self::KEY_ALWAYS_LOADED) {
            $handlers = array_merge($handlers, $this->getHandlers($name));
        }

        $logger = new Logger($name, $handlers);
        Handler::getInstance()->register($logger);
        $this->cache[$name] = $logger;

        return $logger;
    }
```

PHPCI/Logging/Handler.php
```diff
@@ -33,30 +53,39 @@ class Handler
...
     {
-        $this->logger = $logger;
+        $this->logger = $logger ?: new Logger('PhpError');
     }
 
-    /**
-     * Register a new log handler.
+    /** Adds a logger to push logs into.
```
mavimo: Comment should be in the second line.

Adirelle:
@tvbeek: Rebased on latest master.

tvbeek:
I run phpunit and it say everything is oke.
But I get this error ( a couple of times)
`2015-04-22 11:34:43] something.ERROR: ErrorException: 2: posix_isatty(): could not use stream of type 'MEMORY' in phpci/vendor/symfony/console/Symfony/Component/Console/Output/StreamOutput.php line 101 (uncaught exception) at phpci/vendor/symfony/console/Symfony/Component/Console/Output/StreamOutput.php line 101 {"exception":"[object] (ErrorException(code: 0): 2: posix_isatty(): could not use stream of type 'MEMORY' in phpci/vendor/symfony/console/Symfony/Component/Console/Output/StreamOutput.php line 101 at phpci/vendor/symfony/console/Symfony/Component/Console/Output/StreamOutput.php:101)"} []`

Do you know how to fix it?

Adirelle:
I don't even know how to cause it. Do you use a specific loggerConfig.php ?

Adirelle:
Ok, I reproduced it. Working on it.

Adirelle:
After trying to tweak the existing code to fix that "mysterious bug", I've finally decided to let Monolog\ErrorHandler log the errors (as PHPCI is already using Monolog and I expect it to be widely tested) and have the original PHPCI handler throw exceptions on PHP errors.

mavimo:
@Adirelle how you reproduce the error?

Adirelle:
Running PHPUnit in the command line (in a Debian Wheezy box) caused the errors to appear. It disappeared when I enabled process isolation in phpunit.xml. I think it was related to LoggerConfig tests, which didn't remove the error handler after each test.

tvbeek:
@Adirelle Thanks that you fix it :)

@mavimo Did you had a reason to remove the flag:ready-to-merge ?

mavimo:
@tvbeek o removed the flag since I have no time to check and I see a lot of
commits added after flag is attached to PR.

bootstrap.php
```diff
@@ -43,9 +42,10 @@
 // Load Composer autoloader:
 require_once(dirname(__FILE__) . '/vendor/autoload.php');
 
+\PHPCI\ErrorHandler::register();
+
 if (defined('PHPCI_IS_CONSOLE') && PHPCI_IS_CONSOLE) {
     $loggerConfig = LoggerConfig::newFromFile(__DIR__ . "/loggerconfig.php");
```
tvbeek: You have removed the Handler:register call, is the LoggerConfig::newFromFile stil needed because the var $loggerConfig isn't used in this file?

tvbeek:
@mavimo good point, I check the code and found one question for @Adirelle after that I think we can give the flag again and merge it

bootstrap.php
Adirelle: The $loggerConfig isn't used in bootstrap.php but in both the console and daemonise scripts. We can 1) let these scripts as they are, 2) duplicate this piece of code in the two scripts, 3) merge daemonise into console and move this piece of code there.

Adirelle:
@tvbeek @mavimo This seems fair. Actually I was hesitating between closing this PR and submitting a new one or pushing this reworked version.

bootstrap.php
tvbeek: I think for now we can do option 1 and maybe in the future option 3. (Option 2 doesn't sound good for me)

mavimo: I vote for 1 then we can consider a better log management during rework.

mavimo:
@tvbeek @Adirelle  :+1: for this PR. I can't merge it actually anyone can do that?

Issue #884, opened 2015-03-26
Intent to implement: using symfony/process


Adirelle:
What's the idea ?

[http://symfony.com/doc/current/components/process.html symfony/process] 

Why ?

* It is a well tested component.
* It gracefully handles Windows shenanigans.
* It handles process timeouts (both based on total duration and output idling). Both ca
* It can be used to run PHP in process isolation.

dancryer:
I think that this, combined with #885 - would make for a nice step towards a cleaner command executor.

We'd need to do a thorough review before merging, but would be pleased to see a PR implementing these ideas.

I assume you'll keep the `CommandExecutor` classes, but just make them use `Process` internally?

Adirelle:
Yep, to avoid breaking change, the CommandExecutor would stay as a thin layer between existing code and symfony/process.

Adirelle:
So I started working on this (see there ) and achieved these features:
* catches stdout, stderr and exit code as before,
* "live-streams" stdout and stderr to the loggger.
* implements a per-command and global timeout.

However it the second point gives rise to several questions:
* it doesn't have the exit code while streaming stderr to the logger so it cannot chose not to log it or log it at a level depending on the exit code.
* I've seen they are $logOutput, $verbose and $quiet attributes in Builder and BaseCommandExecutor. I haven't verified how they are used but the plugin yet. Are there some guidelines and/or explanations about what they should do and how to use them ?

Right now, I've implemented the following behaviour but I feel this should be discussed in a more general topic about logging.
* CommandExecutor as new method, execute, that accepts a workingDir, a verbosity level and "out" arguments for exitCode, stdout and stderr feedback.
* the verbosity level accepts four values :
 * STEALTHED: logs neither stdout nor stderr,
 * QUIET: logs stdout and stderr as DEBUG; if the commands exits with a non-zero status, stderr is logged as ERROR again,
 * NORMAL: logs stdout as DEBUG, stderr as NOTICE,
 * VERBOSE: logs stdout as INFO, stderr as NOTICE,
* since Builder is still used by all plugins, it implements the compatibilty layer, converting $quiet, $verbose and $logExecOutpout to a verbosity level and "plublishing" stdout and stderr through getLastOutput() and getLastError().

Any thought ?

PR #896, opened 2015-03-30
[WIP] Added ability to configure the executable to all plugins.

LewisW:
Contribution Type: refactor
Primary Area: plugins

Description of change:
Some plugins let you override their executable, others don't. 

Description of solution:
For the sake of consistency and configurability, I've updated them all to allow the overriding of their executable. Ideally I'd like to move this logic to be automatic for each plugin (or usage of findBinary).

# TODO
- [x] Update individual plugins
- [ ] Update documentation
- [ ] Discuss feasibility of moving the executable overriding to `Builder::findBinary`
- [ ] Implement basic tests for each plugin

[Review on Reviewable](https://reviewable.io/reviews/block8/phpci/896)

### DIFF
--- a/PHPCI/Plugin/PhpCsFixer.php
+++ b/PHPCI/Plugin/PhpCsFixer.php
@@ -36,6 +36,8 @@ class PhpCsFixer implements \PHPCI\Plugin
     protected $verbose    = '';
     protected $diff       = '';
     protected $levels     = array('psr0', 'psr1', 'psr2', 'all');
+    protected $workingdir;
+    protected $executable;
 
     /**
      * Standard Constructor
@@ -67,7 +69,7 @@ class PhpCsFixer implements \PHPCI\Plugin
         $curdir = getcwd();
         chdir($this->workingdir);
 
-        $phpcsfixer = $this->phpci->findBinary('php-cs-fixer');
+        $phpcsfixer = $this->executable;
 
         if (!$phpcsfixer) {
             $this->phpci->logFailure(Lang::get('could_not_find', 'php-cs-fixer'));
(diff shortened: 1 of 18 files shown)


mavimo:
I prefer to move in PHPCI\Plugin the executable extraction based on plugin constant, eg:

```php
class PHPCI\Plugin {
  const EXECUTABLE_NAME = '';

  protected $options = [];
  /* ... */

  protected function getExecutable()
  {
      if (isset($this->options['executable'])) {
          return $this->options['executable'];
      }
      $configDirectories[] = '/bin';
      $configDirectories[] = '/sbin';
      $configDirectories[] = '/urs/bin';
      $configDirectories[] = '/urs/sbin';
      $configDirectories[] = getcwd();
      $configDirectories[] = rtrim(getenv('HOME') ?: getenv('USERPROFILE'), '/\\');
      // ...
      
      $locator = new FileLocator($configDirectories);
      return $locator->locate(self::EXECUTABLE_NAME, null, false);
  }
}


class MyPlugin extends 
{
    const EXECUTABLE_NAME = 'myexec';

    public function execute()
    {
      // ...
      $this->getExecutable();
      // ...
    }
}
```

in order to reduce code duplication, aggregate logic, improve code readability.

LewisW:
Agreed, that's a much cleaner solution, although I think I'll hold off for now as #884 might mitigate the need to do this.

PHPCI/Plugin/PhpCsFixer.php
```diff
@@ -88,6 +90,12 @@ public function execute()
      */
     public function buildArgs($options)
```
phpcibot: The method buildArgs() has a Cyclomatic Complexity of 10. The configured cyclomatic complexity threshold is 10.

PR #897, opened 2015-03-30
[WIP] Added clean-up build status on command terminate/interrupt

LewisW:
Contribution Type: cosmetic
Primary Area: builder

Description of change:
Cancelling the run command left a build set as "Running".

Description of solution:
This PR implements signal handling in Unix and will clean-up the status of a build in the event that a command is killed with a SIGINT (pressing ctrl + c) or a SIGTERM (something else telling the command to stop).

# TODO
- [ ] Translations, currently they have the English text there with a TODO, I cannot offer anymore translations
- [ ] The composer dependency is locked to a commit as there is no stable version tagged, I've asked the maintainer to tag a stable version.

### DIFF
--- a/PHPCI/Command/RunCommand.php
+++ b/PHPCI/Command/RunCommand.php
@@ -29,7 +29,7 @@ use PHPCI\Model\Build;
 * @package      PHPCI
 * @subpackage   Console
 */
-class RunCommand extends Command
+class RunCommand extends SignaledCommand
 {
     /**
      * @var OutputInterface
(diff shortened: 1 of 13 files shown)


PHPCI/Command/RunCommand.php
```diff
@@ -29,7 +29,7 @@
 * @package      PHPCI
 * @subpackage   Console
 */
-class RunCommand extends Command
+class RunCommand extends SignaledCommand
```
mavimo: We don't need to import the ```SignaledCommand``` class from ```Che\ConsoleSignals``` namespace?

Adirelle:
Wouldn't a custom signal handler that throws exception (like what the error handle already does) have the same result without the need of another external library ?

LewisW:
Agreed, this library doesn't handle all the situations correctly anyway.

PR #908, opened 2015-04-09
phpci:daemon cleanup

Adirelle:
Reworked the DaemonCommand.
* Accepts options for the PID and log file.
* Uses the best process control method available on the installation : POSIX, windows CLI or Unix CLI.
* Checks the daemon actually started or stopped.
* Try to terminate then kill the daemon.
* Uses either the logger or the output interface.

Notes:
* this PR is just a few commits extracted from #864.
* IMO, the log file should be handled directly in the phpci:daemonise command using LoggerConfig.

### DIFF
--- a/PHPCI/Command/DaemonCommand.php
+++ b/PHPCI/Command/DaemonCommand.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * PHPCI - Continuous Integration for PHP
  *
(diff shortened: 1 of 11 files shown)


mavimo:
Since the last commit is not builded from PHPCI i removed the label, waiting for a new build.

mavimo:
@Adirelle can you do a small change and push again in order to re-run the build? Thx!

Adirelle:
@mavimo ping !

PHPCI/ProcessControl/ProcessControl.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+namespace PHPCI\ProcessControl;
+
+/**
+ * A stateless service to check and kill system processes.
+ *
+ * @author Adirelle <adirelle@gmail.com>
+ */
+interface ProcessControl
```
mavimo: I prefer to have ```ProcessControlInterface``` as name, is possible to change it?

PR #910, opened 2015-04-11
Let CommandExecutor::findBinary throw an exception.

Adirelle:
Since CommandExecutor::findBinary is followed by this piece of code in *all* plugins, I modified CommandExecutor::findBinary to throw an exception, unless the new optional argument $quiet yields true (in which case the method returns null, as before).

```
if(!$binary) {
    $this->phpci->logFailure(Lang::get('could_not_find', 'binary'));
    return false;
}
```

### DIFF
--- a/PHPCI/Helper/BaseCommandExecutor.php
+++ b/PHPCI/Helper/BaseCommandExecutor.php
@@ -9,9 +9,9 @@
 
 namespace PHPCI\Helper;
 
-use \PHPCI\Logging\BuildLogger;
+use Exception;
+use PHPCI\Logging\BuildLogger;
 use Psr\Log\LogLevel;
-use PHPCI\Helper\Lang;
 
 /**
  * Handles running system commands with variables.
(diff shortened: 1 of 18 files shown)


mavimo:
This involve some part that are related to #888 I prefer to postpone this since we don't merge it.

Adirelle:
I should be able to rebase #888 on top of this, if need be.

PHPCI/Helper/BaseCommandExecutor.php
```diff
@@ -161,30 +160,30 @@ public function findBinary($binary, $buildPath = null)
...
-                break;
+                return $findCmdResult;
             }
         }
-        return $binaryPath;
+
+        if ($quiet) {
+            return;
+        }
+        throw new \Exception(\PHPCI\Helper\Lang::get('could_not_find', implode('/', $binary)));
```
mavimo:
```php
use Exception;
use PHPCI\Helper\Lang;

// ...
        throw new Exception(Lang::get('could_not_find', implode('/', $binary)));
// ...
```

Adirelle:
@mavimo Fixed.

PR #915, opened 2015-04-15
Change gitlab file link in plugins

rm3nchaca:
Running builds leave a file link with an error like "http://gitlab.example.com/root/project/blob/master/index.php#L6" but it is pointing to the actual file, not the file with a bug, example "http://gitlab.example.com/root/project/blob/97f0a6453d5913f4b55d660fbf2d629240ca7237/index.php#L6"

### DIFF
--- a/PHPCI/Model/Build/GitlabBuild.php
+++ b/PHPCI/Model/Build/GitlabBuild.php
@@ -47,8 +47,7 @@ class GitlabBuild extends RemoteGitBuild
             'http://%s/%s/blob/%s/{FILE}#L{LINE}',
             $this->getProject()->getAccessInformation("domain"),
             $this->getProject()->getReference(),
-            $this->getgetCommitId()
-            //$this->getBranch()
+            $this->getCommitId()
         );
     }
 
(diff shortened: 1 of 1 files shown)


PHPCI/Model/Build/GitlabBuild.php
```diff
@@ -47,7 +47,8 @@ public function getFileLinkTemplate()
             'http://%s/%s/blob/%s/{FILE}#L{LINE}',
             $this->getProject()->getAccessInformation("domain"),
             $this->getProject()->getReference(),
-            $this->getBranch()
+            $this->getgetCommitId()
```
mavimo: This method seems to be wrong, should be ```getCommitId()```?

PHPCI/Model/Build/GitlabBuild.php
```diff
@@ -47,7 +47,8 @@ public function getFileLinkTemplate()
...
+            //$this->getBranch()
```
mavimo: Remove unrequired comment.

PHPCI/Model/Build/GitlabBuild.php
rm3nchaca: you're right!! Thanks! I edited directly on github, in my local install it's ok..
Saluti dal Messico!

mavimo: Thx!
Saluti dall'Italia :smile:

PR #922, opened 2015-04-17
Allow overriding the build directory in local_vars.php

ronanchilvers:
Hi guys

This is a very simple PR to allow moving the build directory to a custom location using a define in vars.php/local_vars.php. My use case is that I want to be able to store the build on a much faster disk.

I haven't provided any unit tests for this PR. Let me know if I need to. Similarly if there's documentation that I should update somewhere.

Thanks

Ronan

### DIFF
--- a/PHPCI/Builder.php
+++ b/PHPCI/Builder.php
@@ -287,7 +287,7 @@ class Builder implements LoggerAwareInterface
      */
     protected function setupBuild()
     {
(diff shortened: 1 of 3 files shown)


mavimo:
See #834

ronanchilvers:
Apologies - I didn't see the duplicate. I'll cancel this PR.

Issue #924, opened 2015-04-20
Commit link on Manual Build


alasdairdc:
Feature request:
Would it be possible to show the commit hash/version when doing Manual builds?

tvbeek:
This is done in #928 so I close this issue

PR #928, opened 2015-04-21
Commit Id on Manual Build 

Adirelle:
Primary Area: builder
Link to Bug: #924 

Description of change:
*  Show the actual commit hash for manual builds.

Description of solution: 
* In BuildService, sets localized "Manual build" as the build comment for manual builds.
* In RemoteGitBuild::postCloneSetup, update the commit id with the actual hash (using `get rev-parse HEAD`).


mavimo:
:+1:

Issue #907, opened 2015-04-06
Mail notification template


Adirelle:
Would it possible to select the template used for notification using a parameter instead of the build outcome ? I'd like failed builds to send a mail with the full log while successful builds could just send a short notice.

(This could be used on the instance that tests the Github project to send feedback about PRs.)

tvbeek:
You mean like a default template and an option to change it
```
email:
        committer: true
        default_mailto_address: one@example.com
        template: with_logs
```

I think that is a nice feature

PR #933, opened 2015-04-22
Mail template option

Adirelle:
Primary Area: plugins
Link to Bug: #907

Description of change: 
* Add an "template" option for the email plugin to select the template to use (between "short" and "long").

Description of solution:
* Add a common layout for emails, with dynamic parts (color and introducing phrases).
* Rename the old "failed" and "success" templates to "short" and "long".
* For backwards compatibility, when the "template" option is missing, the selection is based on the build status: "short" for failed builds, "long" for successful builds.


PHPCI/Plugin/Email.php
```diff
@@ -70,12 +70,21 @@ public function execute()
 
         $buildStatus  = $this->build->isSuccessful() ? "Passing Build" : "Failing Build";
         $projectName  = $this->build->getProject()->getTitle();
-        $mailTemplate = $this->build->isSuccessful() ? 'Email/success' : 'Email/failed';
+        $mailTemplate = $this->build->isSuccessful() ? 'Email/long' : 'Email/short';
```
tvbeek: Isn't it more logic to get a long email on failure and the short one on success? (Maybe that is just logic for me, I'm not sure)

PHPCI/Plugin/Email.php
```diff
@@ -70,12 +70,21 @@ public function execute()
...
+
+        if (!empty($this->options['template'])) {
```
tvbeek: Is it possible to check if the template exist? (If I made a typo it can result in an error)

Adirelle: Not sure. I don't know if the b8 framework allows this.

PHPCI/Plugin/Email.php
Adirelle: You're right and I agree. I just have kept the current behavior. Should we change that ?

Adirelle:
@mavimo Changes done.

PR #938, opened 2015-04-23
Build removal safeguard

Adirelle:
Contribution Type: bug fix
Primary Area: builder

Description of change:

Fixing a potential bug introduced by #834: when deleting a build with no identifier, `Build::removeBuildDirectory` may try to remove the whole PHPCI/build tree. This should no happen in production, but it does happen in the tests.

Description of solution:

Have `Build::getBuildPath` returns null instead of an incorrect but valid path when the Build does not have an identifier (i.e. hasn't be persisted yet) and have `Build::removeBuildDirectory` double-checks the path.

### DIFF
--- a/PHPCI/Model/Build.php
+++ b/PHPCI/Model/Build.php
@@ -221,10 +221,13 @@ class Build extends BuildBase
     /**
      * Return the path to run this build into.
      *
-     * @return string
+     * @return string|null
      */
     public function getBuildPath()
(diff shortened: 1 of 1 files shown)


tvbeek:
:+1: Thanks, I was already wondering why the .gitignore from the build directory was removed after running phpunit

mavimo:
Closed

PR #942, opened 2015-04-25
Separated strings for 'archived'/'archived_menu'

corpsee:
Contribution Type: bug fix
Primary Area: view layout
Description of change:

Separated strings for 'archived'/'archived_menu'


mavimo:
Changes llok fine, can i ask why you need to have a separate string for archived menu?

corpsee:
Because in russian there are two different words. Word 'archive' (archived_menu in lang file) is for the archived projects and 'archived' is for archived project (**one** project).

mavimo:
So, the main issue is related to one/may element in translation?

corpsee:
Yes. And to be precise it is the different words, but in english they have the equal forms. I think in many non-english languages, these words are not equal too.

corpsee:
@mavimo I answered your question?

PR #951, opened 2015-04-28
Fail-safe rendering of plugins in build view.

Adirelle:
**Contribution Type:** bug fix
**Primary Area:** front-end

**Description of change:**
Prevent the whole display to fail if an exception occurs in one plugin.

**Description of solution:**
Use try...catch in ActiveBuild.renderPlugin and log the exception to the JS console.

### DIFF
--- a/public/assets/js/build.js
+++ b/public/assets/js/build.js
@@ -70,9 +70,8 @@ var Build = Class.extend({
     },
 
     renderPlugins: function() {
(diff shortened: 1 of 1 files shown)


dancryer:
Hey @Adirelle - Please could you check this still works on the latest version of PHPCI? If so, :+1:  from me.

PR #952, opened 2015-04-28
Improved output log handler

Adirelle:
**Contribution Type:** new feature
**Primary Area:** logger

**Description of change:**
- Filter messages according to the verbosity level of the OutputInterface; this honors the -v, -vv, -vvv and -q options. The messages are filtered this way, by increasing verbosity:
  - -q: errors and higher,
  - (nothing): warnings and higher; **this means the default display is far less verbose than the current one**,
  - -v: notices and higher,
  - -vv: infos and higher,
  - -vvv: all messages.
- Send messages of severity ERROR or higher to stderr.
- Display line breaks inside the messages.
- Display the stack trace of exceptions.
- Severity-based coloring: 
  - red for ERROR or higher, 
  - yellow for WARNING, 
  - green for NOTICE,
  - else white.

**Description of solution:**

Modification of the OutputLogHandler class.


dancryer:
Hi @Adirelle - As with the other PR, please can you verify that this still works with the current PHPCI version? If so, :+1:  from me.

PR #971, opened 2015-05-11
Copyplugin wipe fix

strietzel:
Moved "*" wildcard character out of the quotes, because within the quotes it gets escaped and is not used as wildcard anymore.

### DIFF
--- a/PHPCI/Plugin/CopyBuild.php
+++ b/PHPCI/Plugin/CopyBuild.php
@@ -26,7 +26,6 @@ class CopyBuild implements \PHPCI\Plugin
     protected $wipe;
     protected $phpci;
     protected $build;
-    protected $symlink;
 
     /**
      * Set up the plugin, configure options, etc.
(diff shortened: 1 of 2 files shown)


PHPCI/Plugin/CopyBuild.php
```diff
@@ -75,7 +75,7 @@ public function execute()
     protected function wipeExistingDirectory()
     {
         if ($this->wipe === true && $this->directory != '/' && is_dir($this->directory)) {
-            $cmd = 'rm -Rf "%s*"';
+            $cmd = 'rm -Rf "%s"*';
```
mavimo: Since there is ``ìs_dir``` chech are we sure we can manage directory where path is not resolved?
```
php > var_dump(is_dir('/tm'));
bool(false)

php > var_dump(is_dir('/tmp'));
bool(true)
```
I suppose the ```*``` is not supported.

strietzel: for explanation:
in my opinion the whole topic about the wildcard is as follows:
The original command rm -Rf "%s*" did not work, because the %s is escaped and not interpreted as wildcard.
I recognized, that even if the wildcard would be interpreted correctly, $this->directory NEEDS a trailing slash to wipe all content inside the specified directory.
So I fixed the command to work properly by moving te wildcard out of the quotes.

I think the final step is to make sure, that $this->directory always has a trailing slash.

What do you think?

Issue #960, opened 2015-05-01
No results shown after a complete build


Henk8:
After a complete build the results are not shown in the detail page of a commit. There are no errors and when I run the plugins manualy everything is fine.

The tests that are performed are:
- Codeception
- Copy paste detector
- PHPLoc
- DocBlockParser

The logs are correctly stored in the database, but not shown.

Build log:
```
Working copy created: /
RUNNING PLUGIN: php_loc
phploc 2.0.6 by Sebastian Bergmann.
Size
  Lines of Code (LOC)                               91
  Comment Lines of Code (CLOC)                      44 (48.35%)
  Non-Comment Lines of Code (NCLOC)                 47 (51.65%)
  Logical Lines of Code (LLOC)                      13 (14.29%)
    Classes                                         11 (84.62%)
      Average Class Length                          11
      Average Method Length                          1
    Functions                                        0 (0.00%)
      Average Function Length                        0
    Not in classes or functions                      2 (15.38%)
Complexity
  Cyclomatic Complexity / LLOC                    0.00
  Cyclomatic Complexity / Number of Methods       1.00
Dependencies
  Global Accesses                                    0
    Global Constants                                 0 (0.00%)
    Global Variables                                 0 (0.00%)
    Super-Global Variables                           0 (0.00%)
  Attribute Accesses                                 9
    Non-Static                                       9 (100.00%)
    Static                                           0 (0.00%)
  Method Calls                                       9
    Non-Static                                       9 (100.00%)
    Static                                           0 (0.00%)
Structure
  Namespaces                                         0
  Interfaces                                         0
  Traits                                             0
  Classes                                            1
    Abstract Classes                                 0 (0.00%)
    Concrete Classes                                 1 (100.00%)
  Methods                                            9
    Scope
      Non-Static Methods                             9 (100.00%)
      Static Methods                                 0 (0.00%)
    Visibility
      Public Methods                                 7 (77.78%)
      Non-Public Methods                             2 (22.22%)
  Functions                                          0
    Named Functions                                  0 (0.00%)
    Anonymous Functions                              0 (0.00%)
  Constants                                          0
    Global Constants                                 0 (0.00%)
    Class Constants                                  0 (0.00%)
PLUGIN: SUCCESS
BUILD SUCCESS
Removing Build.
```

![noresults](https://cloud.githubusercontent.com/assets/11833224/7431500/5dd134ec-f01f-11e4-93af-9db5cdd8eb9c.png)

[...]

Issue #929, opened 2015-04-21
Error Log notice - property of non-object


alasdairdc:
Currently receiving the php error log message:

    PHP Notice:  Trying to get property of non-object in /var/www/PHPCI/PHPCI/Application.php on line 107
relating to the following line of code:
```php
if ($this->response->hasLayout() && $this->controller->layout) {
```
using the default Debian PHP settings.

Issue #719, opened 2014-12-18
[BUG] Wrong controller action


zilionis:
I enter link:
http://phpci.block8.net/register

I see result:

Fatal error: Class 'PHPCI\\Controller' not found in /www/phpci/vendor/block8/b8framework/b8/Application.php on line 93

Exprected result:
Page 404

dancryer:
http://phpci.block8.net/register is not a valid link. There is no /register in PHPCI.

PR #976, opened 2015-05-14
Application::handleRequest now properly displays exceptions.

Adirelle:
Contribution Type: bug fix
Primary Area: front-end
Link to Bug: #960, #929, #719.

Description of change: properly display exceptions happening while handling a request.

Description of solution: modified Application::handleRequest.

### DIFF
--- a/PHPCI/Application.php
+++ b/PHPCI/Application.php
@@ -94,15 +94,21 @@ class Application extends b8\Application
                 $this->response->setContent($this->controller->layout->render());
             }
 
+            return $this->response;
+
+        } catch (\HttpException $ex) {
+            $this->response->setResponseCode($ex->getErrorCode());
         } catch (\Exception $ex) {
-            $this->config->set('page_title', 'Error');
-
-            $view = new View('exception');
-            $view->exception = $ex;
-
-            $this->response->setResponseCode($ex instanceof HttpException ? $ex->getErrorCode() : 500);
-            $this->response->setContent($view->render());
+            $this->response->setResponseCode(500);
         }
+
+        // We only get there if an excetion has been caught
+        $this->config->set('page_title', 'Error');
+
+        $view = new View('exception');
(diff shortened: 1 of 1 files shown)


Adirelle:
Combined with https://github.com/Block8/b8framework/pull/41 (on the 1.x version of the framework), this allows to gracefully handle "not found" exceptions.

PHPCI/Application.php
```diff
@@ -86,31 +86,24 @@ public function handleRequest()
...
-            $this->response->setResponseCode($ex->getErrorCode());
-            $this->response->setContent($view->render());
         } catch (\Exception $ex) {
             $this->config->set('page_title', 'Error');
 
             $view = new View('exception');
             $view->exception = $ex;
 
-            $this->response->setResponseCode(500);
+            $this->response->setResponseCode($ex instanceof HttpException ? $ex->getErrorCode() : 500);
```
mavimo: I'm not sure have a condition check here is good.

Adirelle: This is the only differences between the two catch blocks. If HttpException used/overrode Exception::getCode instead of defining its own getErrorCode, this checks wouldn't even be necessary.

PHPCI/Application.php
```diff
@@ -86,30 +86,29 @@ public function handleRequest()
...
-            $this->response->setContent($view->render());
-        } catch (\Exception $ex) {
-            $this->config->set('page_title', 'Error');
+                $this->controller->layout->content  = $this->response->getContent();
+                $this->response->setContent($this->controller->layout->render());
+            }
 
-            $view = new View('exception');
-            $view->exception = $ex;
+            return $this->respone;
```
mavimo: Typo, should be ```$this->response```(missing S)

mavimo:
This should not merge since the b8 framework fix was added.

PR #979, opened 2015-05-15
Prevents getByLoginOrEmail from being erased by phpci:generate.

Adirelle:
Contribution Type: bug fix
Primary Area: ORM

getByLoginOrEmail has been added to UserStoreBase  in #873. However, this class is generated and overwritten by the phpci:generate command. The solution is to move it into UserStore, which should host such custom methods.


tvbeek:
:+1:

PR #981, opened 2015-05-15
Added call for setOptions, which has been omitted, leading to no opti…

teemule11:
…ons being evaluated


mavimo:
@teemule11 The fix is good, but can you clean the translations strings in order to have a separate MR for translation?

mavimo:
Seems to be a duplicate for !975

PR #987, opened 2015-05-21
Adding an option for Plugin::Composer to pass the no_dev option.

teemule11:
If set to true, it will execute composer with the --no-dev option, which usually suffices for testing in most projects. Default is set to false.

My intention: as composer dependency resolution tends to take quite some time, build speed can be greatly reduced by resolving only those dependencies which are really needed.


teemule11:
Forgot to mention my intention: as composer dependency resolution tends to take quite some time, build speed can be greatly reduced by resolving only those dependencies which are really needed.

teemule11:
Forgot to mention my intention: as composer dependency resolution tends to take quite some time, build speed can be greatly reduced by resolving only those dependencies which are really needed.

Issue #988, opened 2015-05-22
Codeception report failed when Test format is used


tankist:
There is a difference in XML report when using Cest and Test formats in Codeception:

Cest:
```xml
<testcase file="/home/stalin/www/PHPCI/PHPCI/build/2/tests/unit/Service/Search/Adapter/StandardCest.php" name="testStandardAdapterFindUseRelevanceSortAndNoAlgorithmRelevance" class="Unit\Service\Search\Adapter\StandardCest" feature="test standard adapter find use relevance sort and no algorithm relevance" assertions="1" time="0.115426"/>
```

Test:
```xml
<testcase name="testBeforeTestProfiler" class="Unit\Library\Codeception\Extensions\ProfilerTest" file="/home/stalin/www/PHPCI/PHPCI/build/2/tests/unit/Library/Codeception/Extensions/ProfilerTest.php" line="53" assertions="1" time="0.008662"/>
```

As you can see there is no `feature` attribute in Test XML entry

Codeception plugin by default is using `feature` attribute so for Test testcases it will be empty.

zot24:
How this have been addressed? It's fails for me and I'm in the 2.1.5 Codeception version.

Do we need to have both fixes? https://github.com/tankist/PHPCI/commit/ef21502e99893779f5b56cb281186ca210230918 and https://github.com/Codeception/Codeception/pull/1973? if so why this issue haven't been closed?

zot24:
Just for the record my issue here was similar but not the same:

```
Exception: Notice: Undefined index: class in /Users/zot/Desktop/phpci/PHPCI/Plugin/Util/TestResultParsers/Codeception.php line 67
```

Because there is not `class` on my `report.xml` file

PR #989, opened 2015-05-22
Fix for #988 Codeception report failed when Test format is used

tankist:
Added class::method fallback for feature entry to prevent empty report for Codeception plugin

### DIFF
--- a/PHPCI/Plugin/Util/TestResultParsers/Codeception.php
+++ b/PHPCI/Plugin/Util/TestResultParsers/Codeception.php
@@ -62,6 +62,11 @@ class Codeception implements ParserInterface
                     $testresult['class'] = (string) $testcase['class'];
                 }
 
+                // PHPUnit testcases does not have feature field. Use class::method instead
+                if (!$testresult['feature']) {
+                    $testresult['feature'] = sprintf('%s::%s', $testresult['class'], $testresult['name']);
+                }
+
                 if (isset($testcase->failure)) {
                     $testresult['pass'] = false;
                     $testresult['message'] = (string) $testcase->failure;


PHPCI/Plugin/Util/TestResultParsers/Codeception.php
```diff
@@ -62,6 +62,11 @@ public function parse()
                     $testresult['class'] = (string) $testcase['class'];
                 }
 
+                //PHPUnit testcases does not have feature field. Use class::method instead
```
mavimo: Space between ```//``` and comment.

PHPCI/Plugin/Util/TestResultParsers/Codeception.php
```diff
@@ -50,10 +50,10 @@ public function parse()
 
             foreach ($testsuite->testcase as $testcase) {
                 $testresult = array(
-                    'suite' => (string) $testsuite['name'],
+                    'suite' => ucfirst((string) $testsuite['name']),
                     'file' => str_replace($this->phpci->buildPath, '/', (string) $testcase['file']),
                     'name' => (string) $testcase['name'],
-                    'feature' => (string) $testcase['feature'],
+                    'feature' => ucfirst((string) $testcase['feature']),
```
mavimo: Why UCFirst?

PHPCI/Plugin/Util/TestResultParsers/Codeception.php
```diff
@@ -50,10 +50,10 @@ public function parse()
 
             foreach ($testsuite->testcase as $testcase) {
                 $testresult = array(
-                    'suite' => (string) $testsuite['name'],
+                    'suite' => ucfirst((string) $testsuite['name']),
```
mavimo: Why UCFirst?

PHPCI/Plugin/Util/TestResultParsers/Codeception.php
tankist: Just for pretty output. "feature" attribute is taken by Codeception from the method name and in most cases this method name is started from lower case letter.

tankist:
Writing my last comment I have noticed that "feature" attribute is taken from test method name. For my quickfix if "feature" attribute is empty I'm filling it as class::method. So what do you think, @mavimo, maybe it's better to add logic to split method name as well as Codeception do it for Cest classes?

I mean I do in this fix: "Unit\Library\Ras\Logger\Adapter\LimitedFileTest::testCanCreateLimitedFileAdapter"
What can I do: "Test can create limited file adapter"

In second case we will have unified Codeception log for both Cest/Test formats.

PHPCI/Plugin/Util/TestResultParsers/Codeception.php
tankist: Updated

mavimo:
@tankist since you are introducing two different changes (formatting cleanup and failure when Test format is used) i prefer to split this MR in two part. The first one only fix the Test format report (and I think is fine) for the formatting I'm not sure is "good" alter the codexception result since can be "source of issues".

tankist:
So you mean remove "ucfirst" and leave if block only?

PR #991, opened 2015-05-22
Bugfix: Default Codeception xml path "tests/_output/" but now 100%

nigeon:
But now it's 100% configurable.


mavimo:
Older project should need to be re-configured, this is not back-compatible config change.

nigeon:
But this was a bug introduced in the previous revision, that btw, was not back-compatible as the parameter name changed... :s

El 23/05/2015, a les 15:21, Marco Vito Moscaritolo <notifications@github.com> va escriure:

PR #993, opened 2015-05-24
Fixed GithubBuild::getDiffLineNumber method for correct phpcpd work

corpsee:
**Contribution Type:** bugfix
**Primary Area:** phpcpd plugin

**Description of change:**
On processing phpcpd output GithubBuild throw error:
```
Warning: Illegal offset type in .../GithubBuild.php line 224
```


tvbeek:
@corpsee can you check if this is already fixed with 92ee90998a1b6b218656195a78366542d74a97a3 ?

corpsee:
92ee909 doesn't solve my problem, it only hides the error message.

PR #998, opened 2015-05-29
Pdepend Module: make check for Pdepend output dir more robust.

teemule11:
This fixes problems with the xml output directory by the pdfepend module.  Before this fix, the pdepend module would check for "is_writable". But the php function "is_writeable" not only returns false when the directory is not writable, but also if the target directory does not exist at all. The error message given by phpci:pdepend-module is misleading in these cases.
I suggest to a) attempt to create the dir if it does not exist, and b) make the error message more explanatory in case the dir cannot be created or is not accessible.


tvbeek:
:+1:

PR #1002, opened 2015-05-30
Update pimple to version 3.

mavimo:
Update pimple service container to version 3.


Adirelle:
IIRC, pimple 3 has a lot of major changes, including breaking ones.

mavimo:
@Adirelle current usage of Pimple is really limited and seems it work fine (more testing should be useful).

Adirelle:
I think pimple 1 creates new instances of services unless you use the shared helper, whereas pimple 3 creates shared instance by default. This might have some side effects.

Issue #1000, opened 2015-05-29
Confirmation before project is deleted


steverobbins:
The <kbd>Edit Project</kbd> and <kbd>Delete Project</kbd> button are located right next to each other.  I accidentally clicked delete when trying to edit my project and now it's gone.  There should be some sort of confirmation before the project and all it's build history are removed.

ViGo5190:
https://github.com/Block8/PHPCI/issues/739

PR #1005, opened 2015-06-02
Confirmation dialog before project is deleted (Fixes #739, #1000)

### LINKED ISSUE #739: Confirmation for deleting project
I think it would be great, if we'll add confirmation for deleting project. Now it can be deleted by random click...

steverobbins:


steverobbins:
Why was this rejected?

REBELinBLUE:
Hmm yeah @dancryer why have you rejected it? Seems rather minor change which could be quite important

steverobbins:
It looks like the delete link (among others) was removed? https://github.com/Block8/PHPCI/blob/master/PHPCI/View/layout.phtml#L229

Or maybe it moved somewhere else.

REBELinBLUE:
Ah yes of course, it is now a button at the top of the project details

PR #1006, opened 2015-06-02
Fix run command

ViGo5190:
Fix checking if  some project is building now.

in `PHPCI/Command/RunCommand.php`

we have some check 
```
if (!$this->isFromDaemon && in_array($build->getProjectId(), $running)) {
```

but in `$running` we have smtg like that:
```
array(1) {
  [5]=>
  bool(true)
}
```

So we need change `in_array` to `array_key_exists `


mavimo:
can you squash all your commit into a single one? 9 commits for a single line of code is a bit too much :)

Issue #1007, opened 2015-06-02
Fatal: Cannot add project


th3fallen:
When adding a project with github application configured, after filling out all data i get this... 

Sorry, there was a problem
SQLSTATE[HY000]: General error: 1364 Field 'access_information' doesn't have a default value

mrjgreen:
I seem to have had this issue too. I basically fixed it by taking the NOT NULLs off all text fields and removing NOT NULL and adding DEFAULT NULL to the datetime fields and adding DEFAULT '' to access_information and last_commit.

On `project`:
    SQLSTATE[HY000]: General error: 1364 Field 'access_information' doesn't have a default value
    SQLSTATE[HY000]: General error: 1364 Field 'last_commit' doesn't have a default value

On `build`:
    SQLSTATE[HY000]: General error: 1364 Field 'log' doesn't have a default value
    SQLSTATE[HY000]: General error: 1364 Field 'started' doesn't have a default value
    SQLSTATE[HY000]: General error: 1364 Field 'finished' doesn't have a default value
    SQLSTATE[HY000]: General error: 1364 Field 'commit_message' doesn't have a default value
    SQLSTATE[HY000]: General error: 1364 Field 'extra' doesn't have a default value

th3fallen:
i did the same @mrjgreen i just wanted to make it known that this is an issue... not sure how that made it into a stable release considering the installer doesn't even work....

Very disappointing.

mavimo:
@th3fallen @mrjgreen can you investigate how to solve the issue on PHP side without change the DB definition? I suppose was an issue on "filled value" on the project & build objects.. thanks!

dodiaraculus:
Seems to be a mysql/php type error.
<pre>
SQLSTATE[HY000]: General error: 1364 Field 'log' doesn't have a default value
SQLSTATE[HY000]: General error: 1364 Field 'started' doesn't have a default value
SQLSTATE[HY000]: General error: 1364 Field 'finished' doesn't have a default value
SQLSTATE[HY000]: General error: 1364 Field 'commit_message' doesn't have a default value
SQLSTATE[HY000]: General error: 1364 Field 'extra' doesn't have a default value
</pre>
Have you tried setting a default value in mysql? I would recommend that you use phpmyadmin.

curtisdf:
[...]

Issue #977, opened 2015-05-14
SQL error during installation


phackwer:
I tried to install PHPCI. During install I've got this error:

Setting up your database... 

                                                                                                             
  [PDOException]                                                                                             
  SQLSTATE[42000]: Syntax error or access violation: 1101 BLOB/TEXT column 'log' can't have a default value  
                                                                                                             
Aparently, there's a error on your SQL script.

So, when I tried to login, I've got all kind of issues.

Tried to reinstall it fixing your sql script. Couldn't find it. :-D So, tried to check how you build the database. Found phinx. made a mitake looking for sphinx. Tried again to look for phinx. Found it. Learned something new. So, I'm trying to see if it's a PHPCI bug or Phinx bug.

Anyway, you should fix it. Problably didn't happen only for me.

phackwer:
Found the bug: your code.

fix_database_columns.php

<?php

use Phinx\Migration\AbstractMigration;

class FixDatabaseColumns extends AbstractMigration
{
    /**
     * Migrate Up.
     */
    public function up()
    {
        $dbAdapter = $this->getAdapter();

        if ($dbAdapter instanceof \Phinx\Db\Adapter\PdoAdapter) {
            $pdo = $dbAdapter->getConnection();
            $pdo->exec('SET foreign_key_checks = 0');
        }

        $build = $this->table('build');
        $build->changeColumn('project_id', 'integer', array('null' => false));
        $build->changeColumn('commit_id', 'string', array('limit' => 50, 'null' => false));
        $build->changeColumn('status', 'integer', array('null' => false));
        $build->changeColumn('log', 'text', array('null' => true));
        $build->changeColumn('branch', 'string', array('limit' => 50, 'null' => false, 'default' => 'master'));
        $build->changeColumn('created', 'datetime', array('null' => true));
        $build->changeColumn('started', 'datetime', array('null' => true));
        $build->changeColumn('finished', 'datetime', array('null' => true));
        $build->changeColumn('committer_email', 'string', array('limit' => 512, 'null' => true));
        $build->changeColumn('commit_message', 'text', array('null' => true));
        $build->changeColumn('extra', 'text', array('null' => true));

        $buildMeta = $this->table('build_meta');
        $buildMeta->changeColumn('project_id', 'integer', array('null' => false));
        $buildMeta->changeColumn('build_id', 'integer', array('null' => false));
        $buildMeta->changeColumn('meta_key', 'string', array('limit' => 250, 'null' => false));
[...]

PR #1008, opened 2015-06-03
resolving log default value issue in migration

th3fallen:
will resolve https://github.com/Block8/PHPCI/issues/1007 https://github.com/Block8/PHPCI/issues/977

### DIFF
--- a/PHPCI/Migrations/20140730143702_fix_database_columns.php
+++ b/PHPCI/Migrations/20140730143702_fix_database_columns.php
@@ -20,7 +20,7 @@ class FixDatabaseColumns extends AbstractMigration
         $build->changeColumn('project_id', 'integer', array('null' => false));
         $build->changeColumn('commit_id', 'string', array('limit' => 50, 'null' => false));
         $build->changeColumn('status', 'integer', array('null' => false));
-        $build->changeColumn('log', 'text', array('null' => true, 'default' => ''));
+        $build->changeColumn('log', 'text', array('null' => true));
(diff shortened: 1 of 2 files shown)


th3fallen:
not big on editing migrations once their completed but these simply will not run as they stand

mavimo:
Migration will be always additive, we can't remove already runned migration code.

th3fallen:
that's impossible then without removing those migrations completely then @mavimo the two migrations that where modified completely fail. unless you have another way you handle it. if so please advise.

mavimo:
@th3fallen create a new migration using:
```
vendor/bin/phinx create MigrationName
```
the fill the methods with appropriate changes.

NB: I'm not sure remove DB constrains is the right solution, are you sure should not be fixed by set the appropriate value on the build/project entity before save it on DB?

th3fallen:
> @th3fallen create a new migration using:

I tried that the issue is the `fix_database_columns` migration fails and blocks all further migrations from running.

th3fallen:
any thoughts @mavimo ?

mikebronner:
I agree with @th3fallen that at one point there were broken migrations out there that just did not work, and you could not upgrade. 

I just did a fresh install and did not encounter those issues anymore. @th3fallen are you still experiencing the issue?

th3fallen:
I moved to something else... but i'll do another install monday and report back.

th3fallen:
Worked for me.

mikk150:
does not work for me, I installed mysql 5.6.27

Issue #833, opened 2015-02-26
[Bug] unknown level "all" in php-cs-fixer


wandersonwhcr:
Olá pessoas (e outros)!

If we don't configure a `level` option in `php_cs_fixer`, the plugin will use an unknown `--level=all` parameter, unrecognized by `php-cs-fixer` binary (v1.4.2).

`
  [InvalidArgumentException]       
  The level "all" is not defined.
`

mikebronner:
Confirmed, this should be an easy fix. If I don't get a PR within the next few days submitted, please feel free to do so.

PHPCI/Plugin/PhpCsFixer.php should probably read:

```php
    protected $workingDir = '';
    protected $level      = ' --level=psr2';
    protected $verbose    = '';
    protected $diff       = '';
    protected $levels     = array('psr0', 'psr1', 'psr2', 'symfony');
```

I will update the wiki in the mean time.

PR #1032, opened 2015-07-09
Update PhpCsFixer plugin

EmmanuelVella:
Update the PhpCsFixer plugin to work with the latest version of php-cs-fixer.

Also added the `--dry-run` option as it does not need to apply the change in the build directory.

This fixes https://github.com/Block8/PHPCI/issues/833

### DIFF
--- a/PHPCI/Plugin/PhpCsFixer.php
+++ b/PHPCI/Plugin/PhpCsFixer.php
@@ -32,10 +32,10 @@ class PhpCsFixer implements \PHPCI\Plugin
     protected $build;
 
     protected $workingDir = '';
-    protected $level      = ' --level=all';
+    protected $level      = ' --level=psr2';
     protected $verbose    = '';
     protected $diff       = '';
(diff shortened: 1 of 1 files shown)


dancryer:
What would be the purpose of running PHP CS Fixer if it doesn't do the fixing?

EmmanuelVella:
It is not necessary to write the file changes in the build directory (they won't be committed), the dry-run is enough to do the report and let the developper fix the CS himself, right ?

dancryer:
I think some people actually do use this to make the changes, and then use either CopyBuild, PackageBuild or the Git plugin with the fixed files.

EmmanuelVella:
Ok, I just removed the dry-run option ;)

EmmanuelVella:
Thanks !

fnash:
@dancryer 

> I think some people actually do use this to make the changes, and then use either CopyBuild, PackageBuild or the Git plugin with the fixed files.

Okay if people fix source files before running tests, so why running php-cs-fixer in builds?

Please can you make this option available? We want to be free to use the tool like as it is.

PR #1035, opened 2015-07-13
Fix missing path in technical debt plugin

EmmanuelVella:
The technical debt plugin is always executed on the root dir. This PR fixes this.

### DIFF
--- a/PHPCI/Plugin/TechnicalDebt.php
+++ b/PHPCI/Plugin/TechnicalDebt.php
@@ -145,7 +145,7 @@ class TechnicalDebt implements PHPCI\Plugin, PHPCI\ZeroConfigPlugin
      */
     public function getErrorList()
     {
-        $dirIterator = new \RecursiveDirectoryIterator($this->directory);
+        $dirIterator = new \RecursiveDirectoryIterator($this->directory . DIRECTORY_SEPARATOR . $this->path);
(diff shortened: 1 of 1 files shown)


mikebronner:
@EmmanuelVella  I don't think this is the right place for the fix -- the setOptions() method is responsible for setting the directory and path, the fix should be there. Overall, the distinction between `directory` and `path` in the plugin seems redundant.

My recommendation would be clean up the setOptions() method and add the path to the directory there, if it is set.

Also, the PR needs to be brought up-to-date with the master branch. 

Please let me know if you aren't able to tackle this, and I'll take it on.

Thanks :)

EmmanuelVella:
Hey @mikebronner, thank you for your feedback.

Sorry for the delay ; i'm quite busy at the moment, please feel free to handle this !

leonboot:
@mikebronner I disagree. When compared to, for instance the CodeSniffer plugin, this plugin constructs the path to scan using the buildPath variable, appended by what was set in the path variable by setOptions(). This plugin handles the setting of the path variable properly, but does not use it in any way. The fix proposed by @EmmanuelVella actually does, in the same fashion it is done in other plugins.

REBELinBLUE:
@jfi You wrote this plugin didn't you?

PR #1040, opened 2015-07-27
Reflow the plugin boxes

micheh:
Contribution Type: bug fix/cosmetic
Primary Area: front-end

Description: The not used plugin boxes simply hide their content, but are included in the layout (because only the content is hidden, not the box itself). Therefore the remaining visible boxes can not flow properly.
Description of solution: The id now is assigned to the box itself, not only the content. This way the hidden boxes are excluded from the layout and the visible boxes can catch up. See below for a comparison.

**Before**

before

**After**

after

### DIFF
--- a/public/assets/js/build.js
+++ b/public/assets/js/build.js
@@ -62,7 +62,7 @@ var Build = Class.extend({
     storePluginOrder: function () {
         var renderOrder = [];
 
-        $('.ui-plugin > div').each(function() {
+        $('.ui-plugin').each(function() {
             renderOrder.push($(this).attr('id'));
         });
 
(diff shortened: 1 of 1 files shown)


micheh:
Previously the boxes also had quite an offset while dragging. I added the `cursorAt` parameter to the drag function, to ensure the boxes always have the same offset while dragging.

AlexStansfield:
Fairly new to phpci, just setting it up and testing at the moment with thoughts to rolling it out for all our projects. One of the issues I have is with the build page layout. 

I just wondered why you closed this PR without a merge. Is this being fixed in another PR?

micheh:
@AlexStansfield PR was closed because it looks like the project is abandoned.

AlexStansfield:
Hmm, yeah. I have to say I'm getting that feeling having taken a look around the github issues and pull requests. It's a shame :(

PR #1053, opened 2015-09-05
Display accurate build status at any point

robertboloc:
Currently, after pushing a new commit if a user asks for the status image, the displayed status will be either failing or passing, based on the previous commit, even if the current build has not completed. 

If the status image is linked somewhere (we use a chrome plugin to embed this image automatically on all our Bitbucket pull requests) most of the time the build is shown as passing, until the actual build completes, and the image is updated.

This causes issues as the build takes a while to complete, and developers have to manually check the PHPCI interface for any running builds before actually trusting the image. I think this defeats the purpose of the badge.

This pull request adds 2 more statuses for the badge (not the build):
- unknown (build is created but is not being executed yet), light grey color
- running (build is being executed by the daemon or cron task), orange color

This displays the actual state of a branch at any point, avoiding to have to manually check for any running builds

### DIFF
--- a/PHPCI/Controller/BuildStatusController.php
+++ b/PHPCI/Controller/BuildStatusController.php
@@ -50,17 +50,30 @@ class BuildStatusController extends \PHPCI\Controller
         $branch = $this->getParam('branch', 'master');
         try {
             $project = $this->projectStore->getById($projectId);
-            $status = 'passing';
+            $status = 'unknown';
 
             if (!$project->getAllowPublicStatus()) {
                 return null;
             }
 
             if (isset($project) && $project instanceof Project) {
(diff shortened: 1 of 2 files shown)


dancryer:
Happy to merge this - but I'd prefer it if the badge used the same set of statuses we use in PHPCI itself... Pending (blue), Running (yellow), Success (green), Failed (red)

robertboloc:
Thanks for the feedback :) I have updated the code to use the same statuses and colours as the rest of the application.

I have also rebased the branch to the latest master.

mikebronner:
:+1: after rebasing again :)

robertboloc:
Rebased :+1:

mikebronner:
:+1:

PR #1063, opened 2015-09-29
Fix failing phpci test

zilionis:
Have wrong formatting.


dancryer:
Sorry, I've only just seen this after I committed the same fix!

PR #1066, opened 2015-10-04
Adding support to check for .phpci.yml so the file can be 'hidden'

REBELinBLUE:
A lot of other systems such as Travis, StyleCI and Codeclimate use a dotfile to hide to config, so I've tweaked the build process to check for a dotfile if phpci.yml doesn't exist


REBELinBLUE:
@dancryer Not sure if you'll like this one, but i'll bug you on hangouts sometime over the week ;)

PR #1067, opened 2015-10-05
Adding support for a beanstalkd-based queue

dancryer:


REBELinBLUE:
jesus, that was quick. Will install it later and try it on my server :D

PR #1077, opened 2015-10-10
Fix Arrangement of Build Result Components

mikebronner:
This will allow the other non-hidden components to properly reflow on the page.


dancryer:
I assume you've tested that this change doesn't break anything else? :)

mikebronner:
I'm using it on my local build without any issue so far. Of course I don't have ALL plugins running, but it works for the ones I do have running.

dancryer:
Good enough for me.

PR #1082, opened 2015-10-13
Add the ability to hide the slack attachment

REBELinBLUE:
The attachment may not be useful in situations where you customise the message such as using a message in success, failure, fixed or broken rather than complete. For instance you may specifically mention "the build is broken!" so you don't need the attachment repeating it

### DIFF
--- a/PHPCI/Plugin/SlackNotify.php
+++ b/PHPCI/Plugin/SlackNotify.php
@@ -24,6 +24,7 @@ class SlackNotify implements \PHPCI\Plugin
     private $username;
     private $message;
     private $icon;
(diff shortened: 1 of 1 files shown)


PHPCI/Plugin/SlackNotify.php
```diff
@@ -116,12 +99,40 @@ public function execute()
             $message->setIcon($this->icon);
         }
 
-        $message->attach($attachment);
+        // Include an attachment which shows the status and hide the message
```
phpcibot: PHPCS: Expected 1 newline after opening brace; 2 found

PR #1081, opened 2015-10-13
Added 2 new build stages

REBELinBLUE:
"Broken" happens when the build fails after previously being successful

"Fixed" happens when the build succeeds after having previously failed

PR #1083, opened 2015-10-13
Added 2 new build stages

REBELinBLUE:
"Broken" happens when the build fails after previously being successful

"Fixed" happens when the build succeeds after having previously failed

Clean up of #1081


PHPCI/Model/Project.php
```diff
@@ -131,4 +154,4 @@ public function getIcon()
 
         return $icon;
     }
-}
```
phpcibot: PHPCS: Expected 1 newline at end of file; 0 found

PHPCI/Builder.php
```diff
@@ -402,4 +420,4 @@ function () use ($self) {
 
         return $pluginFactory;
     }
-}
```
phpcibot: PHPCS: Expected 1 newline at end of file; 0 found

phpcibot: PHPCS: Expected 1 newline at end of file; 0 found

mikebronner:
Does the build class need to be updated with these two new statuses as well? Wondering if that's causing a disconnect for other plugins, like the Slack notification plugin.

```php
class Build extends BuildBase
{
    const STATUS_NEW = 0;
    const STATUS_RUNNING = 1;
    const STATUS_SUCCESS = 2;
    const STATUS_FAILED = 3;
```

REBELinBLUE:
It doesn't actually have the status of the build, the build is still either STATUS_FAILED or STATUS_SUCCESS, it is just that it checks when the build finishes whether or not the previous build was the opposite result and if so runs the options defined

dancryer:
^^ What he said.

mikebronner:
Thanks :) that's what I thought as well, just found it strange that the Slack issue was raised today right after this update. :) (Was thinking there might be a connection.) I haven't had issues with slack notifications either. Working fine for me.

PR #1087, opened 2015-10-13
Fixing javascript error when trying to delete a project

steverobbins:
I am receiving a javascript error when trying to delete a project:

> Uncaught SyntaxError: missing ) after argument list

It seems it is missing a closing `)` on the function call.


steverobbins:
See PR 1086

PR #1097, opened 2015-10-15
Fixes the technical debt migration

REBELinBLUE:
The technical debt migration was wrong

@dancryer I wasn't sure whether to fix it in the existing migration, because it is safe to rerun, or to add a new one


mikebronner:
:+1:

mikebronner:
Does this also address #1035 ?

PR #1111, opened 2015-10-27
ImageOptim'ed assets/

PierreLvx:
Lossless optimisation of all img assets. Quick win.

optim


mikebronner:
:+1:

REBELinBLUE:
:+1: I think @pornel would actually be annoyed with me for having not thought to do this myself >_<

pornel:
:)

PR #1118, opened 2015-11-06
CakePHP 2.x helpers and spanish updating

tranfuga25s:


slevesque:
Are you using it actually and how is it working with unit test in cakephp 2.x ?

tranfuga25s:
This is used on our work to test an CakePHP 2.x application with Codeception.

This helps to configure all the files that are specific for the enviroment where must be runned and leaves all set to use the testing with codeception.

There is no integration for unit testing yet.

PR #1123, opened 2015-11-12
Giving plugins an isAllowedInStage() method

### LINKED ISSUE #1090: Slack notifications always failed
Hi

I'm using the `Slack Notifier` plugin and every time a build runs and completes, no matter what status, the message in my Slack channel is always:

`Status: Failed`

Does anyone know why this is?

### LINKED ISSUE #1122: Have plugins throw an error if they are run at a stage when they shouldn't be run
I.e. When the email, slack and hipchat plugins are run in `test`

Probably somewhere like `executePlugins()` or `doExecutePlugins()` in `PHPCI/Plugin/Util/Executer.php` calling something like a `canRun` method on the plugin, passing in the stage and the plugin returning true or false

Unfortunately Plugins just implement an interface rather than extend an abstract class so we can't just add a `canRun` method which just returned true, so maybe have the `executePlugins()` method do a `method_exists` on the the Plugin, if so call it otherwise just assume the plugin can run; that way we don't need to define the method on all plugins which just returns true in most cases.

@dancryer ?

dancryer:
Giving plugins an isAllowedInStage() method to protect them from being run at inappropriate times.

Fixes #1122
Fixes #1090

### DIFF
--- a/PHPCI/Model/Build.php
+++ b/PHPCI/Model/Build.php
@@ -168,7 +168,7 @@ class Build extends BuildBase
             }
 
             foreach (array('setup', 'test', 'complete', 'success', 'failure') as $stage) {
-                if ($className::canExecute($stage, $builder, $this)) {
+                if ($className::canRunZeroConfig($stage, $builder, $this)) {
                     $config[$stage][$className] = array(
                         'zero_config' => true
                     );
--- a/PHPCI/Plugin.php
+++ b/PHPCI/Plugin.php
@@ -15,5 +15,6 @@ namespace PHPCI;
 */
 interface Plugin
 {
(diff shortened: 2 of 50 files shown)


dancryer:
@REBELinBLUE Please can you sense check this before I merge it?

REBELinBLUE:
Ha ha, that was quick! Seems right to me, can't see anything wrong

REBELinBLUE:
I've just noticed, some of the plugins already have a `canExecute` method, such as PHPCS

```php
    public static function canExecute($stage, Builder $builder, Build $build)
    {
        if ($stage == 'test') {
            return true;
        }

        return false;
    }
```

shouldn't that me used?

PR #1128, opened 2015-11-14
Added branch and update_only parameters

### LINKED ISSUE #1127: Intent to implement: Branch support for Deployer plugin
Hoping to do this over the weekend, I am going to update the plugin to pass in the branch and an `update_only` option.

Also need to update Deployer obviously, the branch will mean Deployer will deploy the supplied branch and `update_only` means it will only deploy if the current deployed branch matches

REBELinBLUE:
Fixes #1127


dancryer:
Just so you'll shush.

REBELinBLUE:
:-1: Only took 10 days of bugging you :p Anyone would think you had a life

PR #1129, opened 2015-11-16
New "Source" Parameters for the CopyBuild Plugin

GregJPreece:
## Overview

Modifying the PHPCI CopyBuild plugin to add additional options, which allow the user to specify a subdirectory of the build folder to copy from, rather than copying over the entire build. Additionally, it is now possible to copy only the build's contents, without including the source directory itself, by setting the `include_src_dir` flag to `false`.

**Note:** The `directory` parameter has been renamed to `dest` to avoid confusion when specifying a source directory with `src`. The old `directory` parameter will continue to work under the new code, for backward compatibility.
## Examples

Here's an example of how the copy plugin was defined previously. (This is still valid.)

```
success:
    copy_build:
        directory: "/path/to/test.domain.com/"
        wipe: true
        respect_ignore: true
```

This copies the build directory and everything in it, so if your build directory is called `34_abcdef`, then the path of your built software will be `/path/to/test.domain.com/34_abcdef`. 

As an example, you might have run a Phing build using PHPCI, and you want the `target/` subdirectory from that build to be copied, rather than the root of the build. The new `src` parameter solves this problem:

```
success:
    copy_build:
        src: "target/"
        dest: "/path/to/test.domain.com/"
        wipe: true
        respect_ignore: true
```

This will only copy the specified folder to the destination directory. In this case, the built software would now be at `/path/to/test.domain.com/target/`.

However, you also might not want that `target/` subdirectory to come along with the software. The new flag, `include_src_dir`, allows users to get around this issue:

```
success:
    copy_build:
        src: "target/"
        dest: "/path/to/test.domain.com/"
        wipe: true
        respect_ignore: true
        include_src_dir: false
```

In this case, the contents of the source  folder (or the build folder, if a source is not specified) will be copied, leaving the source folder itself behind, and the path of the built software will be `/path/to/test.domain.com/`.

_Please note:_ While I've done my best to verify the command syntax on Windows using a VM (as you can see from my second commit) I don't have a running PHPCI Windows instance to test this live, though I believe the edit is correct and non-harmful.


PHPCI/Plugin/CopyBuild.php
```diff
@@ -35,12 +36,22 @@ class CopyBuild implements \PHPCI\Plugin
      */
     public function __construct(Builder $phpci, Build $build, array $options = array())
```
phpcibot: The method __construct() has an NPath complexity of 9375. The configured NPath complexity threshold is 200.

PHPCI/Plugin/CopyBuild.php
```diff
@@ -48,20 +59,32 @@ public function __construct(Builder $phpci, Build $build, array $options = array
...
 
-        if ($this->directory == $build) {
+        if ($this->dest == $build) {
             return false;
         }
 
         $this->wipeExistingDirectory();
 
-        $cmd = 'mkdir -p "%s" && cp -R "%s" "%s"';
+        $suffix = '';
```
phpcibot: PHPCS: Expected 1 newline after opening brace; 2 found

PHPCI/Plugin/CopyBuild.php
GregJPreece: This is just reading in the parameters and setting defaults if they haven't been set, hence the high NPath complexity. In reality I can't see this being an issue.

GregJPreece:
I've had this pull request open for months, and haven't seen any human input (I'm guessing because of the bot complaints). Apparently a build is failing and I can't see why. I'd like to get this resolved - am I missing a step in becoming a contributor? What is currently wrong with the code that prevents it from building?

GregJPreece:
Found this, seems I haven't followed the contributor process: https://github.com/Block8/PHPCI/wiki/Contributing-to-PHPCI - the changes here are fairly small. Do I need to drop the pull request and go back around?

PR #1154, opened 2016-01-05
Update German translation.

staudenmeir:
Contribution Type: improvement
Primary Area: front-end

Description of change: Updated German translation.


staudenmeir:
I just noticed that the other commit (949bfc5) was also added to the pull request. Is there a way to remove it? How can I prevent this from happening?

PR #1159, opened 2016-01-08
Update English translation.

staudenmeir:
Contribution Type: improvement
Primary Area: front-end

Description of change: Removed duplicate keys.


funivan:
`      'shell_not_enabled' => 'The shell plugin is not enabled. Please enable it via config.yml.', ` 
This line probably is invalid. We can enable shell plugin via `local_vars.php` but not `config.yml`

Issue #1135, opened 2015-11-23
Branch name limited to 50 characters


martinvenus:
Hi,

is there any reason to limit branch name to 50 characters?

Table build, column branch - VARCHAR(50)

Is it possible to extend column branch to at least 100 characters?

martinvenus:
Hi,

is there any plan to solve this problem in 1.8? We have long branch name very often and in that case phpci is not able to run test.

Thank you very much.

PR #1192, opened 2016-03-10
Branch name in build table extended to 250 characters

martinvenus:
Sometimes it happened to me that I have longer branch name and PHPCI then cannot run tests (because of 50 characters limit on branch name).


martinvenus:
It solves issue #1135

martinvenus:
Is there any problem please?

PR #1239, opened 2016-06-29
Fix 'CommitterEmail must be a string' err in bitbucket webhook

bochkovprivate:
Contribution Type: bug fix
Link to Intent to Implement:
Link to Bug:

- [x] Front-End
- [ ] Builder
- [ ] Build Plugins

- [x] I have checked that another pull request for this purpose does not exist.
- [x] Do the PHPCI tests pass?

I've noticed that bitbucket fails with message "CommitterEmail must be a string" while trying to push payload to phpci webhook. WebhookController.php was assuming that commit's author email is always like some@mail.com, while in real life email can be without "<>" symbols.

### DIFF
--- a/PHPCI/Controller/WebhookController.php
+++ b/PHPCI/Controller/WebhookController.php
@@ -116,8 +116,11 @@ class WebhookController extends \b8\Controller
         foreach ($payload['push']['changes'] as $commit) {
             try {
                 $email = $commit['new']['target']['author']['raw'];
-                $email = substr($email, 0, strpos($email, '>'));
-                $email = substr($email, strpos($email, '<') + 1);
+                if (strpos($email, '>') !== false) {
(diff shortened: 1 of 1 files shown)


PHPCI/Controller/WebhookController.php
```diff
@@ -116,8 +116,11 @@ protected function bitbucketWebhook($payload, $project)
         foreach ($payload['push']['changes'] as $commit) {
             try {
                 $email = $commit['new']['target']['author']['raw'];
-                $email = substr($email, 0, strpos($email, '>'));
```
phpcibot: PHPCS: TRUE, FALSE and NULL must be lowercase; expected "false" but found "FALSE"

bochkovprivate: FIxed.

PR #1253, opened 2016-08-27
Read standard output and standard error of child process using stream…

TobiasKappe:
Contribution Type: bug fix
Link to Intent to Implement: none
Link to Bug: none

- [ ] Front-End
- [x] Builder
- [ ] Build Plugins

- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have created or updated the relevant documentation for this change on the PHPCI Wiki.
- [x] Do the PHPCI tests pass?

I noticed that the test suite of one of our projects (over 700 tests) would hang while running the PHPUnit plugin. This problem was later traced back to the fact that, if a process fills up the buffer of the standard error pipe before closing the standard output pipe, it hangs. As a result, the call to stream_get_contents for the standard output pipe does not return, as the standard error pipe needs to be read from in order for the process to continue. This change uses stream_select to read from both pipes whenever data becomes available.

I'm also including a unit test that tests whether a CommandExecutor is capable of handling a lot (i.e., more than the typical pipe buffer size of 64kb) of output on standard error before output on standard output is given.

### DIFF
--- a/PHPCI/Helper/BaseCommandExecutor.php
+++ b/PHPCI/Helper/BaseCommandExecutor.php
@@ -92,14 +92,14 @@ abstract class BaseCommandExecutor implements CommandExecutor
         $pipes = array();
         $process = proc_open($command, $descriptorSpec, $pipes, $this->buildPath, null);
 
+        $this->lastOutput = '';
+        $this->lastError = '';
+
         if (is_resource($process)) {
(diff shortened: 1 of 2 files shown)


TobiasKappe:
Hi Dan, could you tell me why this PR was closed? As far as I can tell from the source code, the issue still remains.

Issue #1248, opened 2016-08-09
Intent to implement: paratest plugin


bdanchilla:
I would like to add `paratest` support (parallel phpunit testing) https://github.com/brianium/paratest as a new plugin. 

Note: I can currently run paratest within the `phpunit` plugin by changing the binary to 
`$this->phpci->findBinary('paratest');` when I have paratest installed with composer. However, AFAIK the TAP format is not supported by paratest. 

It would be much cleaner to separate out as a new plugin

PR #1270, opened 2016-09-27
Feat/paratest plugin

bdanchilla:
Contribution Type: new plugin
Link to Intent to Implement: https://github.com/Block8/PHPCI/issues/1248

- [ ] Front-End
- [ ] Builder
- [x ] Build Plugins

- [x ] I have read and understood the [contributing guidelines](/.github/CONTRIBUTING.md)?
- [x ] I have checked that another pull request for this purpose does not exist.
- [x ] I have considered, and confirmed that this submission will be valuable to others.
- [ ] Do the PHPCI tests pass?

Plugin for paratest https://packagist.org/packages/brianium/paratest. Paratest runs PHPUnit tests in parallel, speeding up build times. 

As paratest uses PHPUnit, most of the plugin is the same as the existing `php_unit` plugin and it extends the PhpUnit class. 

A consequence of the parallel processing with paratest is that `TAP` results are not supported, as with the php_unit plugin. However, `JUnit` logging is supported and I have added parsing of it after execution to display detailed results.

This is a working first version, but will need some work to make dynamically configurable.

As is, it requires the presence of `vendor/bin/paratest`, the JUnit log file to be written to `junit.log`, the phpunit config file to be `phpunit-parallel.xml` and sets the number of processes to `4`

### DIFF
--- a/PHPCI/Plugin/Paratest.php
+++ b/PHPCI/Plugin/Paratest.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace PHPCI\Plugin;
+
+use PHPCI;
+use PHPCI\Plugin\Util\JUnitParser;
+
+/**
+ * Paratest (parallel PHP Unit) Plugin - Allows parallel PHP Unit testing.
+ * @author       Brian Danchilla
+ * @package      PHPCI
+ * @subpackage   Plugins
+ */
+class Paratest extends PhpUnit
+{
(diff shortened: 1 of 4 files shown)


ptejada:
Your jUnit parser handles skipped and incomplete tests as failures. Ideally there is unittest for this cases.

PR #1322, opened 2017-03-28
Proc Open deadlock

Fenikkusu:
Contribution Type: bug fix
Link to Intent to Implement: N/A
Link to Bug: N/A

* [ ] Front-End
* [x] Builder
* [x] Build Plugins

- [ ] I have checked that another pull request for this purpose does not exist.
- [ ] Do the PHPCI tests pass?

When running on a 7.0 repo, I found that it the phpunit test would hang indefinately. When I ran the command directly in the command line, the command ran but died instantly. After a score of R&D, I finally learned that proc_open has a bug in it where if there is a ton of output on both stdout and stderr at the same time, a dead-lock occurs where one is effectively waiting on the other to finish. This change fixes that converting the processing to be non-blocking and instead relying on PHP to check the status. This bypasses the dead-lock and allows the code to run as expected.

### DIFF
--- a/PHPCI/Helper/BaseCommandExecutor.php
+++ b/PHPCI/Helper/BaseCommandExecutor.php
@@ -102,11 +102,11 @@ abstract class BaseCommandExecutor implements CommandExecutor
             do {
                 usleep(200); //Give It Time To Start
 
-                while(($newData = fread($pipes[1], 1024)) && !feof($pipes[1])) {
+                while (($newData = fread($pipes[1], 1024)) && !feof($pipes[1])) {
                     $lastOutput .= $newData;
                 }
 
-                while(($newData = fread($pipes[2], 1024)) && !feof($pipes[2])) {
+                while (($newData = fread($pipes[2], 1024)) && !feof($pipes[2])) {
                     $lastError .= $newData;
                 }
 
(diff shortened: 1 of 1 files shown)


PHPCI/Helper/BaseCommandExecutor.php
```diff
@@ -94,14 +94,33 @@ public function executeCommand($args = array())
...
             fclose($pipes[0]);
+            stream_set_blocking($pipes[1], 0);
+            stream_set_blocking($pipes[2], 0);
 
-            $this->lastOutput = stream_get_contents($pipes[1]);
-            $this->lastError = stream_get_contents($pipes[2]);
+            $lastOutput = '';
+            $lastError  = '';
+            do {
+                usleep(200); //Give It Time To Start
```
phpcibot: PHPCS: Expected 1 space after WHILE keyword; 0 found

PHPCI/Helper/BaseCommandExecutor.php
```diff
@@ -94,14 +94,33 @@ public function executeCommand($args = array())
...
+
+                while(($newData = fread($pipes[1], 1024)) && !feof($pipes[1])) {
+                    $lastOutput .= $newData;
+                }
```
phpcibot: PHPCS: Expected 1 space after WHILE keyword; 0 found

